PR #1, opened 2018-12-30
Acceptance/integration test environments

lawrencejones:
Testing gets nasty when you're building something like an operator: there are far too many external dependencies and configuring your environment can often be more complex than the tests themselves.

Ideally, we should cover the code with three tiers of testing:
- unit, fast and black box testing where we mock each component
- integration, where we speak to the most critical of dependencies but allow testing of internal package constructs
- acceptance, booting the real thing and validating the happy flow

We know we can achieve unit tests, so integration and acceptance have outstanding questions.

# Integration

Given we're building kubernetes operators, the minimum (and maybe maximum?) we'd want to boot as a dependency for integration testing is a semi-real kubernetes API server. The best experience would allow booting the API server very easily for each suite of tests that requires it, and providing helpers to namespace each tests activity for each teardown.

Luckily the dev experience SIG has developed something for this purpose in `sigs.k8s.io/controller-runtime/pkg/envtest`. This package can be used to boot an API server within about 5s, starting an etcd environment and using external binaries to start the server.

Integration tests should boot the API server and then wrap the reconcile loop to permit monitoring of each reconciliation. Each test would use a fresh manager/controller pair. I'm yet to figure out if separating integration tests into their own package (`pkg/controllers/directoryrolebinding/integration/controller_test.go`) makes sense, or if we lose nothing by placing them within the package they're testing.

I'm also unsure what level we should be testing here. The public interface of our packages is to add the reconcile function to the manager, therefore we should be testing how the manager responds to kubernetes resource creation. What style of test is best for this, though? Do we create long style tests that modify existing resources and see how the manager behaves, or split into many different test/expectations?

# Acceptance

This environment requires a real Kubernetes environment- or something very similar to it- to provide value. We should test only the happy flow here.

This PR includes the creation of an acceptance test environment by using kind (kubernetes in docker) but doesn't include any actual acceptance tests. They're missing because I don't know whether we get any additional value from this, if the integration tests are solid. It's also difficult to test group membership as we'd need to provide credentials to GSuite APIs, and figure out how to manage a 'real' group directory for each test. It's certainly possible but the coverage we get from the integration tests here is pretty good, so as yet of undetermined value...


lawrencejones:
This PR was mainly to keep a record of the work done to achieve this: it can be merged now!

PR #2, opened 2019-01-06
Cache directory lookups, periodically refresh

lawrencejones:
The Google directory lookup APIs are quite expensive, and most users
will want to avoid hitting rate limits in normal operation. Without
caching, the number of API calls will scale linearly with the number of
groups used in all directory role bindings in the cluster: this is
likely to hit limits fast.

This change adds a cached directory role binding so that we cache
results on the group key. Membership changes are infrequent, so we
assume a default cache TTL as 5m. Using the cache means the number of
API requests made to the directory service will scale lineraly with the
number of **distinct** groups used across all directory role bindings in
the cluster, which could be a significant improvement.

Finally, we add a refresh parameter that causes the controller to
periodically reschedule reconciles, ensuring that changes to the group
membership in Google will be propagated to the role bindings in the
cluster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,11 +22,14 @@ package:
 
 To help me personally, my next steps are:
 
-- Modify sudorolebindings to use new controller Add interface
-- Review small packages, consolidate where necessary, add unit tests
-- Document installation procedure
-- Investigate installing webhooks: can we test admission webhooks with the
-  integration test suite?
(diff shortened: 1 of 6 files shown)


lawrencejones:
Tested by setting the refresh interval to 10s, running against a real cluster. Looks good!

PR #6, opened 2019-01-07
Don't use default service account

lawrencejones:
It's best to be explicit when running a pod that it should be targetting
a service account, rather than set default privileges within a
namespace.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -39,6 +39,11 @@ rules:
     verbs:
       - "*"
 ---
+apiVersion: v1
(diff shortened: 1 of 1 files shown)


lawrencejones:
CC @benwheatley

BenWheatley:
I think you've tagged the wrong Ben?

BenWheatley:
(Either that or GitHub has a very silly bug)

lawrencejones:
Hey @benwheatley, this is my bad- I meant to tag @benwh but screwed up. Sorry for the spam!

PR #9, opened 2019-01-14
[skeleton] workloads.lawrjone.xyz (Console)

lawrencejones:
## pkg/apis/workloads: Console (no codegen)

Create the definition of a Console, an instance of a console
environment. This is only the skeleton of what we'll have later, but is
designed to store the User field which we'll then prove can work with
mutating webhooks.

This commit does not update codegen, to make it clearer to reviewers
what has been done by a human and what is machine generated.

## make codegen: add workloads:v1alpha1

Following on from the previous commit that introduced the API
definitions for the workloads API group (and Console resource), this
updates our generated client to include the workloads resources.

## Skeleton workloads controller and webhooks

Create the new workloads manager, along with skeleton controller and
webhook code. This version of the manager uses automatic creation of
webhooks in the cluster, and also manages the service that will point at
the statefulset. Ideally we'd only automatically update the webhook
configuration but the controller-runtime doesn't allow this: we'll
likely change this implementation in future.

## pkg/workloads/console/integration

Create integration tests that verify console behaviour, both the webhook
and the controller together. This commit introduces a number of
integration test helpers that should be really useful for testing
admission webhooks, such as integration.NewServer and similar matchers
for test channels as with the reconcile loop.


lawrencejones:
I'm going to merge this so I can do some smaller cleanups with the code. The PR still serves as a good example of introducing concepts even when closed!

PR #11, opened 2019-01-21
Add stable production deployment

lawrencejones:
For now, use an overlay to represent what we'll deploy to production
that we'll guarantee to be stable.


lawrencejones:
I've used this to deploy the app for now. You could also call this stable, and place the overlay inside of anu for example, but given this is the first place we're using kustomize then maybe we should wait until we have a more mature strategy.

Barring any better suggestions I reckon merging this now makes sense.

PR #14, opened 2019-01-22
Create pod when reconciling consoles

hmac:
This change introduces some real logic to the Consoles controller. It's the first step in fleshing it out, and we'll follow up with many smaller changes. Each commit should be reviewable in isolation. A rough summary of changes:

- Update the Console manifest and add a ConsoleTemplate manifest
- Validate on creation that Consoles have a valid ConsoleTemplateRef
- Create a Job when reconciling a new Console

There are also a number of smaller changes to improve the development workflow.


pkg/integration/integration.go
```diff
@@ -66,7 +66,10 @@ func StartAPIServer(crdDirectoryPath string) (*rest.Config, *envtest.Environment
 // preventing us from integration testing webhooks, we omit this flag and instead enable
 // the AlwaysAdmit plugin which does not clobber the webhook plugin.
 func kubeAPIServerFlags() []string {
-	flags, found := []string{"--enable-admission-plugins=AlwaysAdmit"}, false
+	flags, found := []string{
+		"--disable-admission-plugins=ServiceAccount",
```
lawrencejones: We should leave a comment that explains why this is disabled. Will be good to reference our reasoning so people can update it in future, if necessary.

pkg/workloads/console/controller.go
```diff
@@ -78,6 +82,54 @@ func (r *ConsoleReconciler) Reconcile(request reconcile.Request) (res reconcile.
 		return res, err
 	}
 
-	logger.Log("event", "Reconciled") // TODO
-	return
+	logger.Log("msg", "found console")
```
lawrencejones: Can we get this logging events, and using `logging.WithRecorder`? That way we can debug our controllers by describing the consoles which would be rad. The directoryrolebindings give a good example of this: it's probably clearer to create event constants ([link](https://github.com/lawrencejones/theatre/blob/master/pkg/rbac/directoryrolebinding/controller.go#L30-L38)) and log those than messages, but no harm keeping the messages if we think they're useful.

pkg/workloads/console/controller.go
```diff
@@ -78,6 +82,54 @@ func (r *ConsoleReconciler) Reconcile(request reconcile.Request) (res reconcile.
...
+		Name:      csl.Spec.ConsoleTemplateRef.Name,
+		Namespace: request.NamespacedName.Namespace,
+	}
+	consoleTemplate := &workloadsv1alpha1.ConsoleTemplate{}
+	if err := r.client.Get(r.ctx, consoleTemplateName, consoleTemplate); err != nil {
+		return res, err
+	}
+
+	// Try to find the job for this console
+	logger.Log("msg", "looking for job")
```
lawrencejones: Can we simplify this flow by using `controllerutil.CreateOrUpdate` and just doing that to our job?

pkg/workloads/console/webhook.go
```diff
@@ -63,3 +93,34 @@ func (c *ConsoleAuthenticator) Handle(ctx context.Context, req types.Request) ty
...
+		logger.Log("event", "request.end", "duration", time.Now().Sub(start).Seconds())
+	}(time.Now())
+
+	csl := &workloadsv1alpha1.Console{}
+	if err := c.decoder.Decode(req, csl); err != nil {
+		admission.ErrorResponse(http.StatusBadRequest, err)
+	}
+
+	// validate that the console template ref points to a real console template
+	consoleTemplateName := kubeTypes.NamespacedName{
```
lawrencejones: (As discussed IRL) I think we're going to want to remove this validation, as there's no guarantee when you upload your resource that you're not about to upload your console template, and neither helm nor any kube tooling make guarantees about resource creation ordering.

I think validating this in the controller is better: we're going to have to do this anyway, as just because the template is there when we create it doesn't mean it won't be deleted some other time, and it avoids a coupling between the two resources. We lose the immediacy of feedback but that's more of a Kubernetes API design philosophy rather than something we get to decide.

.circleci/config.yml
```diff
@@ -8,6 +8,25 @@ references:
     working_directory: /go/src/github.com/gocardless/theatre
 
 jobs:
+  check-generated-resources:
```
lawrencejones: 👌

Makefile
```diff
@@ -47,3 +47,8 @@ docker-push:
 
 docker-tag:
 	docker tag $(IMAGE):$$(git rev-parse HEAD) $(IMAGE):latest
+
+# npm install -g prettier
+manifests:
+	controller-gen all
```
lawrencejones: We've not got any instructions on how to install this :( can we pin this using dep and invoke `go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go all` instead? That way we know everyone runs a consistent version.

pkg/workloads/console/integration/integration_test.go
```diff
@@ -66,37 +88,211 @@ var _ = Describe("Console", func() {
 
 	Describe("Creating resources", func() {
 		It("Sets console.spec.user from rbac", func() {
+			consoleTemplate := &workloadsv1alpha1.ConsoleTemplate{
```
lawrencejones: Shall we move this standard console template into a helper function? Something like this: https://github.com/gocardless/theatre/blob/master/pkg/rbac/directoryrolebinding/integration/controller_test.go#L27-L41

It's reused across tests, so I reckon this probably makes things a bit clearer.

lawrencejones: We could even create it in the BeforeEach so we can assume the template is present in each test. Given we're actually testing creation of consoles, rather than the templates, that may make sense. It's what we do for the [admin role](https://github.com/gocardless/theatre/blob/master/pkg/rbac/directoryrolebinding/integration/controller_test.go#L91-L94) in the drb specs.

pkg/workloads/console/integration/integration_test.go
```diff
@@ -66,37 +88,211 @@ var _ = Describe("Console", func() {
...
+
 			csl := &workloadsv1alpha1.Console{
 				ObjectMeta: metav1.ObjectMeta{
-					Name:      "foo",
+					Name:      "my-first-console",
 					Namespace: namespace,
 				},
 				Spec: workloadsv1alpha1.ConsoleSpec{
-					User: "", // deliberately not configured, this should be set by the webhook
+					ConsoleTemplateRef: corev1.LocalObjectReference{Name: "my-first-console-template"},
```
lawrencejones: Nice Kube tekkers. I didn't know this existed when I did the drb, but it totally should be this given it opening the possibility of using apimachinery helpers

👌

pkg/workloads/console/integration/integration_test.go
```diff
@@ -66,37 +88,211 @@ var _ = Describe("Console", func() {
...
-					integration.ReconcileResourceSuccess(namespace, "foo"),
+					integration.ReconcileResourceSuccess(namespace, "my-first-console"),
 				),
 			)
 
 			By("Expect console.spec.user to be set")
 			Expect(csl.Spec.User).To(Equal("system:unsecured"))
 		})
+
+		It("Denies requests with a console template ref pointing to a nonexistent template", func() {
```
lawrencejones: I think we'll want to nix this, given a convergent validation strategy. Going to skip past it for now.

pkg/workloads/console/integration/integration_test.go
```diff
@@ -66,37 +88,211 @@ var _ = Describe("Console", func() {
...
+			By("Expect webhook was invoked")
+			Eventually(valWhCalls, timeout).Should(
+				Receive(
+					integration.HandleResource(namespace, "my-first-console"),
+				),
+			)
+		})
+
+		It("Creates a job", func() {
+			consoleTemplate := &workloadsv1alpha1.ConsoleTemplate{
```
lawrencejones: Not necessarily a good idea, but maybe if we'll relying on this stuff being created we should create and validate it in a BeforeEach, then just have the specs that confirm different bits of behaviour? Like create everything up-front, but then validate the job is there and the roles have been created etc.

pkg/workloads/console/integration/integration_test.go
```diff
@@ -66,37 +88,211 @@ var _ = Describe("Console", func() {
...
+			By("Expect reconcile succeeded")
+			Eventually(calls, timeout).Should(
+				Receive(
+					integration.ReconcileResourceSuccess(namespace, "my-first-pod"),
+				),
+			)
+
+			By("Reconciling again")
+			identifier, _ := client.ObjectKeyFromObject(csl)
+			go func() { consoleController.Reconcile(reconcile.Request{identifier}) }()
```
lawrencejones: What was your thought here on calling Reconcile directly instead of updating the resource and expecting the result via the channel?

pkg/workloads/console/controller.go
benwh: For reference: https://github.com/gocardless/theatre/blob/fbf6d90da93c85395d06f5f05d8a33ccc784d86b/pkg/controllers/sudorolebinding/controller.go#L245-L270

hmac: As discussed - we'll keep on eye on this pattern as we build out the operator. We may need it a lot more for rolebindings.

pkg/workloads/console/integration/integration_test.go
hmac: I had thought that it was nicer to not to a pointless update to the resource just to test reconciliation, but on reflection I think it actually makes more sense to do that, as it's in line with the kind of user action that would cause a reconciliation.

pkg/workloads/console/integration/integration_test.go
hmac: I like this plan - will do

Makefile
hmac: dep is refusing to install this dependency as it "doesn't contain usable Go code". I'm going to move this issue to an Issue to unblock this PR.

hmac: https://github.com/gocardless/theatre/issues/17

pkg/integration/integration.go
hmac: I can't actually remember why we disabled this, other than we got an error of some sort. Do you?

PR #15, opened 2019-01-23
Cleanup of console types

benwh:
- Remove unnecessary protobuf annotations (these aren't supported in CRDs anyway!)
- Add missing comments

### DIFF
--- a/pkg/apis/workloads/v1alpha1/console_types.go
+++ b/pkg/apis/workloads/v1alpha1/console_types.go
@@ -15,7 +15,7 @@ type Console struct {
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
 	Spec   ConsoleSpec   `json:"spec"`
-	Status ConsoleStatus `json:"status,omitempty" protobuf:"bytes,3,opt,name=status"`
+	Status ConsoleStatus `json:"status,omitempty"`
 }
 
 // ConsoleSpec defines the specification for a console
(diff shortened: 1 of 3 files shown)


pkg/apis/workloads/v1alpha1/console_types.go
```diff
@@ -26,44 +26,47 @@ type ConsoleSpec struct {
...
 }
 
+// ConsoleTemplateSpec defines the parameters that a created console will adhere to
 type ConsoleTemplateSpec struct {
-	User                     string                 `json:"user"`
 	Template                 corev1.PodTemplateSpec `json:"template"`
 	DefaultTimeoutSeconds    int                    `json:"defaultTimeoutSeconds"`
 	MaxTimeoutSeconds        int                    `json:"maxTimeoutSeconds"`
-	AdditionalAttachSubjects []rbacv1.Subject       `json:"additionalAttachSubjects"`
+	AdditionalAttachSubjects []rbacv1.Subject       `json:"additionalAttachSubjects,omitempty"`
```
hmac: You thinking we let you omit this field entirely, and just default it to an empty list?

benwh: Yeah, I think that this being optional is a nicer interface, and closer to how the built-in resources do it. Let's give this a go - I'm opening to changing it if it becomes a pain down the road.

PR #16, opened 2019-01-23
Add validation for timeout fields

benwh:
Add sensible maximums for the timeouts on the console and template.

### DIFF
--- a/config/crds/workloads_v1alpha1_console.yaml
+++ b/config/crds/workloads_v1alpha1_console.yaml
@@ -35,14 +35,22 @@ spec:
             reason:
               type: string
             timeoutSeconds:
+              description:
+                Number of seconds that the console should run for. If the
+                process running within the console has not exited before this timeout
+                is reached, then the console will be terminated. If this value exceeds
+                the Maximum Timeout Seconds specified in the ConsoleTemplate that
+                this console refers to, then this timeout will be clamped to that
+                value. Maximum value of 1 week (as per ConsoleTemplate.Spec.MaxTimeoutSeconds).
               format: int64
+              maximum: 604800
+              minimum: 0
               type: integer
             user:
               type: string
           required:
             - user
             - reason
-            - timeoutSeconds
             - consoleTemplateRef
           type: object
         status:
--- a/pkg/apis/workloads/v1alpha1/console_types.go
+++ b/pkg/apis/workloads/v1alpha1/console_types.go
@@ -20,9 +20,20 @@ type Console struct {
 
 // ConsoleSpec defines the specification for a console
 type ConsoleSpec struct {
-	User               string                      `json:"user"`
-	Reason             string                      `json:"reason"`
-	TimeoutSeconds     int                         `json:"timeoutSeconds"`
+	User   string `json:"user"`
+	Reason string `json:"reason"`
+
+	// Number of seconds that the console should run for.
+	// If the process running within the console has not exited before this
+	// timeout is reached, then the console will be terminated.
+	// If this value exceeds the Maximum Timeout Seconds specified in the
+	// ConsoleTemplate that this console refers to, then this timeout will be
+	// clamped to that value.
+	// Maximum value of 1 week (as per ConsoleTemplate.Spec.MaxTimeoutSeconds).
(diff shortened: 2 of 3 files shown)


config/crds/workloads_v1alpha1_console.yaml
```diff
@@ -35,14 +35,21 @@ spec:
             reason:
               type: string
             timeoutSeconds:
+              description:
+                Number of seconds that the console should run for. If the
+                process running within the console has not exited before this timeout
+                is reached, then the console will be terminated. Creation of a Console
+                will be rejected if the specified timeout exceeds the maximum timeout
+                defined within the ConsoleTemplate that the console is referring to.
```
hmac: An alternative approach would be to just clamp the timeout to the max specified in the template. I don't really mind which we do - what do you think?

pkg/apis/workloads/v1alpha1/console_types.go
```diff
@@ -20,9 +20,19 @@ type Console struct {
...
+	Reason string `json:"reason"`
+
+	// Number of seconds that the console should run for.
+	// If the process running within the console has not exited before this
+	// timeout is reached, then the console will be terminated.
+	// Creation of a Console will be rejected if the specified timeout exceeds
+	// the maximum timeout defined within the ConsoleTemplate that the console is
+	// referring to.
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=604800
```
hmac: Worth adding a comment explaining that this is 7 days?

pkg/apis/workloads/v1alpha1/console_types.go
```diff
@@ -52,11 +62,17 @@ type ConsoleTemplate struct {
...
-	AdditionalAttachSubjects []rbacv1.Subject       `json:"additionalAttachSubjects"`
+	User     string                 `json:"user"`
+	Template corev1.PodTemplateSpec `json:"template"`
+
+	// Default time, in seconds, that a Console will be created for
+	// +kubebuilder:validation:Maximum=604800
+	DefaultTimeoutSeconds int `json:"defaultTimeoutSeconds"`
+
+	// Maximum time, in seconds, that a Console can be created for
+	// +kubebuilder:validation:Maximum=604800
```
hmac: Could add minimums on these fields too?

pkg/apis/workloads/v1alpha1/console_types.go
```diff
@@ -20,9 +20,19 @@ type Console struct {
...
+	User   string `json:"user"`
+	Reason string `json:"reason"`
+
+	// Number of seconds that the console should run for.
+	// If the process running within the console has not exited before this
+	// timeout is reached, then the console will be terminated.
+	// Creation of a Console will be rejected if the specified timeout exceeds
+	// the maximum timeout defined within the ConsoleTemplate that the console is
+	// referring to.
+	// +kubebuilder:validation:Minimum=1
```
hmac: Any reason to disallow 0? 1 and 0 both seem pretty silly to me, so I think if we allow the former we should allow the latter.

config/crds/workloads_v1alpha1_console.yaml
benwh: Yes we'll in fact have to take that approach; because the ConsoleTemplate may not be present at the time that the Console is created. Updating this message!

benwh: Well actually we don't *have* to, we could also throw an error in the reconcile loop and mark the console with an error status. But let's go with clamping

pkg/apis/workloads/v1alpha1/console_types.go
benwh: Originally didn't want to have this added to the API docs as it seemed redundant, but fair point.

Added

pkg/apis/workloads/v1alpha1/console_types.go
benwh: Good point, I'll do this. The core k8s types don't appear to use unsigned ints anywhere, so we followed the convention of using plain int on this field, but we definitely don't want clients to submit negative values.

PR #20, opened 2019-01-24
Add the user as a label on the job created by a console

hmac:
This should make it easier to identify the owner of a particular console job.

I really don't like the sanitise function but I don't know a nicer way of expressing that.

### DIFF
--- a/pkg/workloads/console/controller.go
+++ b/pkg/workloads/console/controller.go
@@ -3,6 +3,8 @@ package console
 import (
 	"context"
 	"reflect"
+	"regexp"
+	"strings"
 	"time"
 
 	batchv1 "k8s.io/api/batch/v1"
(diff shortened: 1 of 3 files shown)


pkg/workloads/console/controller.go
```diff
@@ -240,3 +243,21 @@ func findOrCreate(ctx context.Context, c client.Client, obj runtime.Object, name
 	}
 	return err
 }
+
+// sanitiseLabel sanitises a string such that it can be used in a k8s label. Labels must
+// satisfy [A-z0-9-_.]+
+// TODO: is there a neater way to do this?
+func sanitiseLabel(label string) string {
```
lawrencejones:
```go
regexp.MustCompile("[^a-zA-Z\\-_.]").ReplaceAllString("hmac@gocardless.com", "-") => hmac-gocardless.com
```

hmac:
Have gone with that - thanks for the tip!

PR #21, opened 2019-01-25
Use CreateOrUpdate over FindOrCreate

hmac:
This ensures that if we edit any resources, the controller will still
converge them towards the desired state. This is more inline with the
Kubernetes philosophy than "create or leave it alone".

### DIFF
--- a/pkg/workloads/console/controller.go
+++ b/pkg/workloads/console/controller.go
@@ -2,6 +2,7 @@ package console
 
 import (
 	"context"
+	"reflect"
 
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


pkg/workloads/console/controller.go
```diff
@@ -69,174 +73,267 @@ func Add(ctx context.Context, logger kitlog.Logger, mgr manager.Manager, opts ..
...
+type ConsoleReconciler struct {
+	ctx      context.Context
+	logger   kitlog.Logger
+	recorder record.EventRecorder
+	client   client.Client
+	name     types.NamespacedName
+	console  *workloadsv1alpha1.Console
+}
+
+func (r *ConsoleReconciler) Reconcile(request reconcile.Request) (res reconcile.Result, err error) {
```
lawrencejones: request is no longer used

hmac:
Superseded by https://github.com/gocardless/theatre/pull/23

PR #22, opened 2019-01-25
Implement expiration of console jobs

benwh:
All consoles will have an expiration time, as we cannot support running
workloads in the cluster for an unconstrained amount of time: this will
both unnecessarily consume resources when interactive consoles are
detached, and hinder maintenance tasks which requires the eviction of
running pods, such as a node pool upgrade.

All consoles have an expiration time in their specification. Once the
console has been running for the amount of time defined by the timeout
the controller will delete the Job object. This deletion propagates to
the Pod which will be removed too, according to its termination grace
period.

### DIFF
--- a/pkg/integration/integration.go
+++ b/pkg/integration/integration.go
@@ -285,7 +285,6 @@ func ReconcileResourceSuccess(namespace, name string) types.GomegaMatcher {
 	return SatisfyAll(
 		ReconcileResource(namespace, name),
 		ReconcileSuccessfully(),
-		ReconcileNoRetry(),
 	)
 }
 
--- a/pkg/workloads/console/controller.go
+++ b/pkg/workloads/console/controller.go
@@ -2,6 +2,8 @@ package console
 
 import (
 	"context"
+	"reflect"
+	"time"
 
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -30,16 +32,19 @@ import (
 const (
 	EventStart    = "reconcile.start"
 	EventComplete = "reconcile.end"
+	EventRequeued = "reconcile.requeued"
 	EventFound    = "found"
 	EventNotFound = "not_found"
 	EventCreated  = "created"
+	EventExpired  = "expired"
+	EventDeleted  = "deleted"
 	EventError    = "error"
 
 	Job             = "Job"
 	Console         = "Console"
 	ConsoleTemplate = "ConsoleTemplate"
 	Role            = "Role"
-	RoleBinding     = "RoleBindings"
+	RoleBinding     = "RoleBinding"
 )
 
 func Add(ctx context.Context, logger kitlog.Logger, mgr manager.Manager, opts ...func(*controller.Options)) (controller.Controller, error) {
@@ -121,22 +126,44 @@ func (r *ConsoleReconciler) Reconcile(request reconcile.Request) (res reconcile.
 			return res, err
 		}
 
-		logger.Log("event", EventNotFound, "resource", Job)
-		job = buildJob(name, consoleTemplate.Spec.Template)
-		if err := controllerutil.SetControllerReference(csl, job, scheme.Scheme); err != nil {
-			return res, err
+		// Only create a console if it hasn't already expired (and therefore been
+		// deleted)
+		if csl.Status.ExpiryTime == nil {
+			job = buildJob(request.NamespacedName, consoleTemplate.Spec.Template)
+			err = r.client.Create(r.ctx, job)
+			if err != nil {
+				return res, err
+			}
+			logger.Log(
+				"event", EventCreated,
+				"resource", Job,
+				"name", name.Name,
+				"user", csl.Spec.User,
+			)
+
+			// Now that we've spawned a console job we can update the status, as
+			// the expiry and phase will have changed.
+			csl, err = r.updateStatus(csl, job)
+			if err != nil {
+				return res, err
+			}
+
+			// Requeue a reconciliation for when we expect the console expiry to fire
+			res = requeueForExpiration(logger, csl.Status)
+		} else {
+			logger.Log(
+				"event", EventExpired,
+				"resource", Job,
+				"name", name.Name,
+				"msg", "Not creating new job for expired console",
+			)
+			return res, nil
 		}
-		if err = r.client.Create(r.ctx, job); err != nil {
-			return res, err
-		}
-		logger.Log(
-			"event", EventCreated,
-			"resource", Job,
-			"name", job.ObjectMeta.Name,
-			"user", csl.Spec.User,
-		)
+
+		return res, err
 	}
 
+	// The console already exists
 	logger.Log(
 		"event", EventFound,
 		"resource", Job,
@@ -149,8 +176,43 @@ func (r *ConsoleReconciler) Reconcile(request reconcile.Request) (res reconcile.
 		return res, err
 	}
 
+	// Update the status fields in case they're out of sync, or the console spec
+	// has been updated
+	csl, err = r.updateStatus(csl, job)
+	if err != nil {
+		return res, err
+	}
+
+	// Terminate if necessary
+	if isJobExpired(csl) {
+		logger.Log(
+			"event", EventExpired,
+			"kind", Job,
+			"resource", request.NamespacedName,
+		)
+
+		err = r.client.Delete(
+			r.ctx,
+			job,
+			client.PropagationPolicy(metav1.DeletePropagationBackground),
+		)
+		if err != nil {
+			// If we fail to delete then the reconciliation will be retried
+			return res, err
+		}
+
+		logger.Log(
+			"event", EventDeleted,
+			"kind", Job,
+			"resource", request.NamespacedName,
+		)
+	} else {
+		// Reconcile was called too early to expire the job, or the
+		// timeout/expiry was updated
+		res = requeueForExpiration(logger, csl.Status)
+	}
+
 	// TODO:
-	//   Terminate the console if it has expired
 	//   GC the terminated console if necessary
 
 	logger.Log("event", EventComplete)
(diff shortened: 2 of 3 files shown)


pkg/integration/integration.go
```diff
@@ -285,7 +285,6 @@ func ReconcileResourceSuccess(namespace, name string) types.GomegaMatcher {
 	return SatisfyAll(
 		ReconcileResource(namespace, name),
 		ReconcileSuccessfully(),
-		ReconcileNoRetry(),
```
benwh: Here I decided to loosen reconciliation result requirement in all tests.

Many control loops will requeue after processing, so when asserting against the reconcile result it makes more sense to make requiring a 'no retry' result the exception rather than the norm.

This didn't break any existing tests, and I figure that if someone wants this functionality they can create another matcher for it.

pkg/workloads/console/integration/integration_test.go
```diff
@@ -85,12 +90,21 @@ var _ = Describe("Console", func() {
 			HaveOccurred(), "failed to create Console",
 		)
 
-		By("Expect reconcile succeeded")
-		Eventually(calls, timeout).Should(
-			Receive(
-				integration.ReconcileResourceSuccess(namespace, "console-0"),
-			),
-		)
+		waitForSuccessfulReconcile = func(times int) {
```
benwh: This seems a bit ugly, but worked for what I needed. It'd be nice to perhaps make it a bit more general and put it in the `integration` package, but I propose that if the reviewer agrees then we create an issue for that, rather than implement in this PR.

pkg/workloads/console/controller.go
```diff
@@ -121,22 +126,44 @@ func (r *ConsoleReconciler) Reconcile(request reconcile.Request) (res reconcile.
...
 		}
 
-		logger.Log("event", EventNotFound, "resource", Job)
-		job = buildJob(name, consoleTemplate.Spec.Template)
-		if err := controllerutil.SetControllerReference(csl, job, scheme.Scheme); err != nil {
-			return res, err
-		}
-		if err = r.client.Create(r.ctx, job); err != nil {
-			return res, err
+		// Only create a console if it hasn't already expired (and therefore been
```
hmac: `Only create a job`?

pkg/workloads/console/controller.go
```diff
@@ -121,22 +126,44 @@ func (r *ConsoleReconciler) Reconcile(request reconcile.Request) (res reconcile.
...
+			logger.Log(
+				"event", EventCreated,
+				"resource", Job,
+				"name", name.Name,
+				"user", csl.Spec.User,
+			)
+
+			// Now that we've spawned a console job we can update the status, as
+			// the expiry and phase will have changed.
+			csl, err = r.updateStatus(csl, job)
```
hmac: I wonder if it'd be simpler if we `defer` this so it always runs at the end of reconciliation?

pkg/workloads/console/controller.go
```diff
@@ -149,8 +176,43 @@ func (r *ConsoleReconciler) Reconcile(request reconcile.Request) (res reconcile.
...
 
+	// Update the status fields in case they're out of sync, or the console spec
+	// has been updated
+	csl, err = r.updateStatus(csl, job)
+	if err != nil {
+		return res, err
+	}
+
+	// Terminate if necessary
+	if isJobExpired(csl) {
```
hmac: I think this is going to try to delete a job multiple times if, for example, a console expires but then its template is updated. Is that worth worrying about?

pkg/workloads/console/integration/integration_test.go
hmac: If this were rspec we'd write  something like
```go
Eventually(calls, timeout).Should(Receive(...).Exactly(n).Times())
```
I don't think gomega supports that, but it would be a nice way of writing it.

pkg/workloads/console/controller.go
benwh: Fixing this

pkg/workloads/console/controller.go
```diff
@@ -240,3 +302,64 @@ func findOrCreate(ctx context.Context, c client.Client, obj runtime.Object, name
...
+
+	// Run a full Update, rather than UpdateStatus, as we can't guarantee that
+	// the CustomResourceSubresources feature will be available.
+	err := r.client.Update(r.ctx, updatedCsl)
+
+	return updatedCsl, err
+}
+
+func calculateStatus(csl *workloadsv1alpha1.Console, job *batchv1.Job) workloadsv1alpha1.ConsoleStatus {
+	newStatus := csl.DeepCopy().Status
```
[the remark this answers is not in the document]
benwh: The problem with that is that once we've removed the console job we'll no longer be able to set the expiration time. The same will apply for things like pod name, once we add that to the status. I think that these are going to be useful things to keep around in the status.

This is a similar approach that some of the more complex core controllers (e.g. Job) take, so we've got a precedent there at least.

Happy to discuss further if you think this is going to cause issues though!

pkg/workloads/console/controller.go
```diff
@@ -240,3 +302,64 @@ func findOrCreate(ctx context.Context, c client.Client, obj runtime.Object, name
...
+
+	if job != nil {
+		// We want to give the console session *at least* the time specified in the
+		// timeout, therefore base the expiry time on the job creation time, rather
+		// than the console creation time, to take into account any delays in
+		// reconciling the console object.
+		// TODO: We may actually want to use a base of when the Pod entered the
+		// Running phase, as image pull time could be significant in some cases.
```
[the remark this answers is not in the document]
benwh: I think that this is going to depend on some of the work required for the `status.Phase` field, i.e. setting up a watcher on Pods. Are you OK if we leave this as a TODO comment at the moment?

Agreed that for the most part this inconsistency won't matter. The only exception would be for short-lived detached tasks, where someone spins up a console expecting a task to take 1m so they set a 5m timeout, but actually it takes 4m to pull the image and then 2m to actually run, so it gets killed.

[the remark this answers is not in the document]
hmac: 👍 on a TODO

pkg/workloads/console/controller.go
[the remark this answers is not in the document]
hmac: If there's precedent for this in kube, then I'm definitely happy to stick with it for now. We can always come back to this if it causes any issues.

pkg/workloads/console/controller.go
benwh: Not sure if this is possible, as you want to to change the return parameters based on the result of the `updateStatus` call (e.g. if updating the status fails then you want to fail, so that the reconcile gets requeued)

AFAIK you can't change return parameters in a defer func? Maybe there's some pattern that you could apply to achieve this, but unless it's super clean then I'm not sure it's worth it, as I reckon there's only going to be 2 `updateStatus` calls in the reconcile loop. WDYT?

hmac: Ah yeah true. Probably won't work, then

pkg/workloads/console/controller.go
benwh: This wasn't actually happening because a little further up there's an early return which bails out if the job is not found but the console expires.

This is bad though I think - it prevents `updateRoleBindings` and `updateStatus` from being called. So I'm going to rework this.

pkg/workloads/console/controller.go
```diff
@@ -141,50 +147,37 @@ func (r *ConsoleReconciler) Reconcile(request reconcile.Request) (res reconcile.
...
 	// Find or create the role and role bindings
 	if err := r.updateRoleBindings(csl, consoleTemplate, name); err != nil {
 		return res, err
 	}
 
 	// Update the status fields in case they're out of sync, or the console spec
 	// has been updated
-	csl, err = r.updateStatus(csl, job)
-	if err != nil {
+	if csl, err = r.updateStatus(csl, job); err != nil {
```
hmac: Keep in mind that the definition of `csl` here doesn't escape this block, so any further references in this function will be to the original value.

hmac:
This looks good to me. I think there _may_ be some things we can clean up, but it's very hard to tell without considering the whole flow in full. Let's merge this and try integrating it with https://github.com/gocardless/theatre/pull/21 - that should reveal whether there's anything we can simplify.

pkg/workloads/console/controller.go
lawrencejones:
```go
# main.go
package main

import (
	"fmt"

	"github.com/davecgh/go-spew/spew"
)

type obj struct {
	msg string
}

func newObj(msg string) *obj { return &obj{msg} }

func main() {
	o := &obj{"hello"}
	if o = newObj("world"); o != nil {
		fmt.Println("executing if")
	}

	spew.Dump(o)
}
```
```
λ (lab) ~> go run main.go
executing if
(*main.obj)(0xc000078290)({
 msg: (string) (len=5) "world"
})
```

hmac: Ah, that might be right. Really fucking annoying how subtle this thing is.

pkg/workloads/console/controller.go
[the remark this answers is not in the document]
lawrencejones: The Kubernetes philosphy on the status subresource is that it represents real-world up-to-date state that is regenerated on each reconcile. They define this in a big API philosophy doc: https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#spec-and-status

benwh:
@lawrencejones We're also going to delete the console itself, but after a larger interval (e.g. 24h) after the timeout. The reason for this is that we wanted a nicer UX when using `anu consoles list`, so that you can see the previous consoles that you started up and when they finished. If we were to remove the console and child objects upon the expiration timeout then the list would only show currently running consoles.

pkg/workloads/console/controller.go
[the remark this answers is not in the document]
benwh: I'm not sure what we've got here necessarily contradicts that philosophy. Maybe one to talk-over in person though

pkg/workloads/console/controller.go
lawrencejones: You can change return parameters in the defer! Just assigning to a named returned variable will change it for the return of the function.

pkg/workloads/console/controller.go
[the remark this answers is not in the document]
lawrencejones: I don't think you do contradict it, I was just putting that link there for easier reading as it's not very discoverable. Better to be making decisions with the context than guessing!

Issue #24, opened 2019-01-29
E2E tests in acceptance


hmac:
Ideally we want to be able to run full end-to-end tests in the acceptance environment, to ensure that our controllers behave properly in something better resembling the real world.

PR #26, opened 2019-01-29
Acceptance test for consoles

hmac:
Closes https://github.com/gocardless/theatre/issues/24

This now works!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,7 +69,7 @@ jobs:
           command: workspace/bin/acceptance prepare --bin workspace/bin
       - run:
           name: Run acceptance tests
-          command: workspace/bin/acceptance run
+          command: KUBECONFIG="$(kind get kubeconfig-path --name='e2e')" workspace/bin/acceptance run
 
 workflows:
   version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,7 +69,7 @@ jobs:
           command: workspace/bin/acceptance prepare --bin workspace/bin
       - run:
           name: Run acceptance tests
-          command: KUBECONFIG="$(kind get kubeconfig-path --name='e2e')" workspace/bin/acceptance run
+          command: workspace/bin/acceptance run
 
 workflows:
   version: 2
--- a/cmd/acceptance/main.go
+++ b/cmd/acceptance/main.go
@@ -34,7 +34,8 @@ var (
 	destroy     = app.Command("destroy", "Destroys the test Kubernetes cluster and other resources")
 	destroyName = destroy.Flag("name", "Name of Kubernetes context to destroy").Default("e2e").String()
 
-	run = app.Command("run", "Runs the acceptance test suite")
+	run     = app.Command("run", "Runs the acceptance test suite")
+	runName = run.Flag("name", "Name of Kubernetes context to against").Default("e2e").String()
 )
 
 // AcceptanceDockerfile defines the docker instructions used to create an acceptance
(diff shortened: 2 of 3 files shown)


pkg/workloads/console/acceptance/acceptance.go
```diff
@@ -0,0 +1,134 @@
...
+
+	By("Expect a console has been created")
+	_, err = workloads.Consoles(namespace).Get(consoleName, metav1.GetOptions{})
+	Expect(err).NotTo(HaveOccurred(), "could not find console")
+
+	By("Expect a job has been created")
+	Eventually(func() error {
+		_, err = core.BatchV1().Jobs(namespace).Get(consoleName, metav1.GetOptions{})
+		return err
+	}).Should(BeNil(), "could not find job")
```
benwh: `ShouldNot(HaveOccurred())` ?

pkg/workloads/console/acceptance/acceptance.go
```diff
@@ -0,0 +1,134 @@
...
+		Expect(err).NotTo(HaveOccurred(), "could not find console")
+		return console.Status.Phase
+	}).Should(Equal(workloadsv1alpha1.ConsoleRunning))
+
+	By("Expect the console phase eventually changes to Stopped")
+	Eventually(func() workloadsv1alpha1.ConsolePhase {
+		console, err = workloads.Consoles(namespace).Get(consoleName, metav1.GetOptions{})
+		Expect(err).NotTo(HaveOccurred(), "could not find console")
+		return console.Status.Phase
+	}, 7, 1).Should(Equal(workloadsv1alpha1.ConsoleStopped))
```
benwh: These random integers threw me a bit at first. Might be clearer to express as `time.Duration`

cmd/acceptance/main.go
```diff
@@ -144,12 +145,16 @@ func main() {
...
-		RunSpecs(new(testing.T), "theatre/cmd/acceptance")
+		if RunSpecs(new(testing.T), "theatre/cmd/acceptance") {
+			os.Exit(0)
+		} else {
+			os.Exit(1)
+		}
 	}
 }
 
 var _ = Specify("Acceptance", func() {
```
benwh: This feels like it should be `Describe("Consoles", ...)` to me

pkg/workloads/console/acceptance/acceptance.go
```diff
@@ -0,0 +1,134 @@
...
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+)
+
+const namespace = "default"
+const consoleName = "console-0"
+
+func Run(logger kitlog.Logger) {
```
benwh: This big method full of a bunch of assertions feels a bit off to me; would a more idiomatic Ginkgo approach be to split it up into a smaller set of `Describe` and `It`s?
This is just me thinking out loud though - maybe there's a good reason it has to be like this. Perhaps we just wait until we need to add another 'group' of acceptance assertions before thinking about this too much.

hmac: I did this on purpose, because I think of this as a single run through of the entire flow. I think splitting it up would make it more difficult to follow.

I'm interested to see how we'd evolve this when we add more tests, but for now I do think this is the clearest way to write it.

.circleci/config.yml
```diff
@@ -69,7 +69,7 @@ jobs:
           command: workspace/bin/acceptance prepare --bin workspace/bin
       - run:
           name: Run acceptance tests
-          command: workspace/bin/acceptance run
+          command: KUBECONFIG="$(kind get kubeconfig-path --name='e2e')" workspace/bin/acceptance run
```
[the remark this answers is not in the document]
hmac: I reckon so!

cmd/acceptance/main.go
hmac: Ah - so actually I disagree. This is the entrypoint for the acceptance tests for the entire theatre system. We may also put DRB tests in here, as well as anything new we add after that. I've added a Describe to the acceptance spec to mark that more clearly.

hmac:
Fixes made - will  merge this once it goes green.

PR #27, opened 2019-01-30
Use Job's ActiveDeadlineSeconds to model console expiration

hmac:
This means we don't have to write any of the expiration logic ourselves.

### DIFF
--- a/pkg/workloads/console/controller.go
+++ b/pkg/workloads/console/controller.go
@@ -135,29 +135,10 @@ func (r *reconciler) Reconcile() (res k8rec.Result, err error) {
 		return res, err
 	}
 
-	// Try to find the job
-	job := &batchv1.Job{}
-	err = r.client.Get(r.ctx, r.name, job)
-	if err != nil && !errors.IsNotFound(err) {
+	// Create or update the job
+	job := buildJob(r.name, r.console, consoleTemplate.Spec.Template)
+	if err := r.createOrUpdate(job, Job, jobDiff); err != nil {
 		return res, err
-	}
-
-	jobExists := err != nil && !errors.IsNotFound(err)
-
-	// If the job hasn't expired, ensure it exists and is up to date
-	if !isJobExpired(r.console) {
-		job = buildJob(r.name, consoleTemplate.Spec.Template)
-		if err := r.createOrUpdate(job, Job, jobDiff); err != nil {
-			return res, err
-		}
-		jobExists = true
-	} else {
-		r.logger.Log(
-			"event", EventExpired,
-			"resource", Job,
-			"name", r.name.Name,
-			"msg", "Not creating new job for expired console",
-		)
 	}
 
 	// Create or update the role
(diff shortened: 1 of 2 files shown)


pkg/workloads/console/integration/integration_test.go
```diff
@@ -216,68 +217,6 @@ var _ = Describe("Console", func() {
...
-
-			job := &batchv1.Job{}
-			identifier, _ = client.ObjectKeyFromObject(csl)
-			err = mgr.GetClient().Get(context.TODO(), identifier, job)
-
-			statusError, _ := err.(*errors.StatusError)
-			Expect(err).To(HaveOccurred(), "the job should be deleted, but it still exists")
-			Expect(err).To(BeAssignableToTypeOf(statusError), "error should be an API error")
-			Expect(statusError.Status()).To(MatchFields(IgnoreExtras, Fields{"Code": BeEquivalentTo(404)}))
-		})
```
hmac: We no longer delete the job, so this test can go.

pkg/workloads/console/integration/integration_test.go
```diff
@@ -216,68 +217,6 @@ var _ = Describe("Console", func() {
...
-
-			waitForSuccessfulReconcile(1)
-
-			job = &batchv1.Job{}
-			identifier, _ = client.ObjectKeyFromObject(csl)
-			err = mgr.GetClient().Get(context.TODO(), identifier, job)
-
-			Expect(err).To(HaveOccurred(), "the job should remain deleted, but it has been recreated")
-			Expect(errors.IsNotFound(err)).To(Equal(true), "a NotFound error was expected, but got a different error")
-		})
```
hmac: The point of this test was to ensure that we only create a job once for every console. Since jobs will now live for the lifetime of the console, this can't happen (since we'd get a resource name clash). We could try testing that updating the timeout of a console after the job's pod has been killed doesn't recreate the pod, but that would be testing the job controller - and we can't do that anyway in this environment, because the job controller isn't running.

pkg/workloads/console/controller.go
```diff
@@ -331,19 +274,15 @@ func calculateStatus(csl *workloadsv1alpha1.Console, job *batchv1.Job) workloads
...
 	existing := existingObj.(*batchv1.Job)
 
-	if !reflect.DeepEqual(expected.Spec.Template, existing.Spec.Template) {
-		// k8s manages the job's metadata, and doesn't allow us to clobber some of the values
-		// it has set (for example, the controller-uid label). To avoid this we only update
-		// the pod template spec.
+	// k8s manages the job's metadata, and doesn't allow us to clobber some of the values
+	// it has set (for example, the controller-uid label). To avoid this we only update
+	// the pod template spec.
+	if !reflect.DeepEqual(expected.Spec.Template.Spec, existing.Spec.Template.Spec) {
```
hmac: The idea here is to reduce the cases where we update the job unnecessarily.

hmac:
@benwh I think this is ready for review! When the acceptance tests are working I'll add some to better test some of this behaviour (e.g. the pod gets created and times out properly).

PR #28, opened 2019-01-30
Add PodName and Phase to Console.Status

hmac:
I've tested this locally and I think it works fine. The nature of these attributes makes them hard or impossible to test in integration (where we have no other resource controllers). I'd like to merge this to unblock the rest of the work and follow up with acceptance tests for this stuff.

### DIFF
--- a/pkg/apis/workloads/v1alpha1/console_types.go
+++ b/pkg/apis/workloads/v1alpha1/console_types.go
@@ -41,8 +41,20 @@ type ConsoleSpec struct {
 type ConsoleStatus struct {
 	PodName    string       `json:"podName"`
 	ExpiryTime *metav1.Time `json:"expiryTime,omitempty"`
-	Phase      string       `json:"phase"`
+	Phase      ConsolePhase `json:"phase"`
 }
+
+type ConsolePhase string
+
+// These are valid phases for a console
+const (
+	// ConsolePending means the console has been created but its pod is not yet ready
+	ConsolePending ConsolePhase = "Pending"
+	// ConsoleRunning means the pod has started and is running
+	ConsoleRunning ConsolePhase = "Running"
+	// ConsoleStopped means the console has completed or timed out
+	ConsoleStopped ConsolePhase = "Stopped"
+)
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
--- a/pkg/workloads/console/controller.go
+++ b/pkg/workloads/console/controller.go
@@ -90,9 +90,9 @@ func (c *Controller) Reconcile(request k8rec.Request) (k8rec.Result, error) {
 		if errors.IsNotFound(err) {
 			logger.Log("event", EventNotFound, "resource", Console)
 		}
-		// If we can't find the console, no meaningful reconciliation we can do. For example,
-		// the console may have been deleted. We don't want to retry in this case, as we'll be
-		// retrying forever. So just return a successful reconcile result.
+		// If we can't find the console, there's no meaningful reconciliation we can do. For
+		// example, the console may have been deleted. We don't want to retry in this case, as
+		// we'll be retrying forever. So just return a successful reconcile result.
 		return k8rec.Result{}, nil
 	}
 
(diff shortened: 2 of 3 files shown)


pkg/workloads/console/controller.go
```diff
@@ -206,6 +229,75 @@ func (r *reconciler) updateStatus(job *batchv1.Job) error {
...
+	// If the pod exists and is running, then the console is running
+	if pod != nil && pod.Status.Phase == corev1.PodRunning {
+		return workloadsv1alpha1.ConsoleRunning
+	}
+
+	// Otherwise, assume the console is pending (i.e. still starting up)
+	return workloadsv1alpha1.ConsolePending
+}
+
+func requeueForExpiration(logger kitlog.Logger, status workloadsv1alpha1.ConsoleStatus) k8rec.Result {
```
hmac: Added this back from the previous version because we want to update the status when the job expires. We don't (currently) watch for events on our child resources so we won't otherwise hear about this.

pkg/workloads/console/controller.go
```diff
@@ -206,6 +226,74 @@ func (r *reconciler) updateStatus(job *batchv1.Job) error {
...
+	// Requeue after the expiry is hit. Add a second to be on the safe side,
+	// ensuring that we'll always re-reconcile *after* the expiry time has been
+	// hit (even if the clock drifts), as metav1.Time only has second-resolution.
+	requeueTime := status.ExpiryTime.Time.Add(time.Second)
+	sleepDuration := requeueTime.Sub(time.Now())
+
+	res := k8rec.Result{}
+	res.RequeueAfter = sleepDuration
+
+	logger.Log(
```
benwh: This can probably be collapsed down to a single line now

pkg/workloads/console/controller.go
```diff
@@ -180,12 +188,24 @@ func (r *reconciler) createOrUpdate(expected reconcile.ObjWithMeta, kind string,
...
 
 func (r *reconciler) updateStatus(job *batchv1.Job) error {
-	newStatus := calculateStatus(r.console, job)
+	// Fetch the job's pod
+	podList := &corev1.PodList{}
+	pod := &corev1.Pod{}
+	opts := client.
+		InNamespace(r.name.Namespace).
+		MatchingLabels(map[string]string{"job-name": job.ObjectMeta.Name})
+	r.client.List(r.ctx, opts, podList)
```
benwh: We're missing a check on the error result here

pkg/apis/workloads/v1alpha1/console_types.go
```diff
@@ -41,9 +41,21 @@ type ConsoleSpec struct {
...
 
+type ConsolePhase string
+
+// These are valid phases for a console
+const (
+	// ConsolePending means the console has been created but its pod is not yet ready
+	ConsolePending ConsolePhase = "Pending"
+	// ConsoleRunning means the pod has started and is running
+	ConsoleRunning ConsolePhase = "Running"
+	// ConsoleStopped means the console has reached its timeout
```
benwh: This comment isn't strictly true; it could also have been that the pod completed before the job timeout

pkg/workloads/console/controller.go
```diff
@@ -206,6 +226,74 @@ func (r *reconciler) updateStatus(job *batchv1.Job) error {
...
+
+	return newStatus
+}
+
+func calculatePhase(job *batchv1.Job, pod *corev1.Pod) workloadsv1alpha1.ConsolePhase {
+	// Currently a job can only have two conditions: Complete and Failed
+	// Both indicate that the console has stopped
+	for _, c := range job.Status.Conditions {
+		if c.Type == batchv1.JobComplete || c.Type == batchv1.JobFailed {
+			return workloadsv1alpha1.ConsoleStopped
```
benwh: I think in the long term we'll want to set a different phase depending on whether the job completed or failed,  to provide a nicer client-side UX. But for now this should get us going

pkg/apis/workloads/v1alpha1/console_types.go
hmac: True!

pkg/workloads/console/controller.go
hmac: Good spot! Will fix.

PR #29, opened 2019-01-30
Clamp console timeout to [1, template.MaxTimeoutSeconds]

hmac:
If the timeout is less than 1, we'll set it to the template's default timeout.
If it's greater than the template's max timeout, we'll set it to the template's max timeout.

If we have reason for wanting consoles with a timeout of zero, we can safely remove the first clause as the openapi validations prevent anyone setting a negative timeout.

### DIFF
--- a/pkg/workloads/console/controller.go
+++ b/pkg/workloads/console/controller.go
@@ -143,8 +143,13 @@ func (r *reconciler) Reconcile() (res k8rec.Result, err error) {
 		return res, err
 	}
 
+	// Clamp the console timeout
+	if err := r.clampTimeout(consoleTemplate); err != nil {
+		return res, err
+	}
+
 	// Create or update the job
-	job := buildJob(r.name, r.console, consoleTemplate.Spec.Template)
+	job := buildJob(r.name, r.console, consoleTemplate)
 	if err := r.createOrUpdate(job, Job, jobDiff); err != nil {
 		return res, err
 	}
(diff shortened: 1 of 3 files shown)


benwh:
Rebased on master to kill the conflicts

pkg/workloads/console/controller.go
```diff
@@ -213,6 +218,32 @@ func (r *reconciler) createOrUpdate(expected reconcile.ObjWithMeta, kind string,
...
 }
 
+// Ensure the console timeout is between [1, template.MaxTimeoutSeconds]
+func (r *reconciler) clampTimeout(template *workloadsv1alpha1.ConsoleTemplate) error {
+	var timeout int
+	switch {
+	case r.console.Spec.TimeoutSeconds < 1:
+		timeout = template.Spec.DefaultTimeoutSeconds
+	case r.console.Spec.TimeoutSeconds > template.Spec.MaxTimeoutSeconds:
+		timeout = template.Spec.MaxTimeoutSeconds
```
benwh: When decreasing the specified timeout to the max timeout, it would be good to emit a warning event, so that we can make the user aware. Otherwise we're just silently mutating what was submitted.

But let's tackle this when we come to sort out events!

pkg/workloads/console/controller.go
```diff
@@ -213,6 +218,32 @@ func (r *reconciler) createOrUpdate(expected reconcile.ObjWithMeta, kind string,
 	return nil
 }
 
+// Ensure the console timeout is between [1, template.MaxTimeoutSeconds]
```
benwh: This isn't *entirely* true. `template.Spec.DefaultTimeoutSeconds` can be 0; in which case we're actually setting the timeout to 0.

Perhaps it's worth changing the validation on this field to be a minimum of 1?

benwh: But maybe setting the min to 1 would break some assumptions we have in the tests...

pkg/workloads/console/integration/integration_test.go
```diff
@@ -102,16 +105,56 @@ var _ = Describe("Console", func() {
...
-
-		waitForSuccessfulReconcile(2)
 	})
 
 	AfterEach(func() {
 		cancel()
 		teardown()
 	})
 
+	Describe("enforcing valid timeout values", func() {
```
benwh: We could potentially add another test case that ensures that if a console is created with a timeout that's non-zero and lower than the max, then it isn't altered.

Could be overkill. WDYT?

pkg/workloads/console/controller.go
hmac: I'd rather just update this comment - I think it's unlikely anyone is going to care whether the timeout is 0 or 1.

pkg/workloads/console/integration/integration_test.go
hmac: Sure, happy to add that.

PR #30, opened 2019-01-31
Make the template own its consoles

hmac:
This means if you delete a console template, all the consoles that
reference it will also be deleted.

There are other ways to handle this issue - for example the controller
could detect a missing template and try to delete the console - but I
think this is the simplest option and will work for most use cases.

### DIFF
--- a/pkg/workloads/console/controller.go
+++ b/pkg/workloads/console/controller.go
@@ -137,6 +137,12 @@ func (r *reconciler) Reconcile() (res k8rec.Result, err error) {
 		return res, err
 	}
 
+	// Set the template as owner of the console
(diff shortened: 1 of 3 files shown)


pkg/workloads/console/controller.go
```diff
@@ -137,6 +137,10 @@ func (r *reconciler) Reconcile() (res k8rec.Result, err error) {
 		return res, err
 	}
 
+	// Set the template as owner of the console
+	// This means the console will be deleted if the template is deleted
+	r.SetConsoleOwner(consoleTemplate)
```
benwh: error result isn't being checked, so this can silently fail 😢

PR #31, opened 2019-01-31
GC consoles after 24 hours

hmac:
I think this does the right thing, but there are currently no tests for it. I'm somewhat hesitant to DI a clock to test this, unless there's a nice pattern for it. Keen to hear advice on what the best approach would be!

### DIFF
--- a/cmd/acceptance/main.go
+++ b/cmd/acceptance/main.go
@@ -34,8 +34,9 @@ var (
 	destroy     = app.Command("destroy", "Destroys the test Kubernetes cluster and other resources")
 	destroyName = destroy.Flag("name", "Name of Kubernetes context to destroy").Default("e2e").String()
 
-	run     = app.Command("run", "Runs the acceptance test suite")
-	runName = run.Flag("name", "Name of Kubernetes context to against").Default("e2e").String()
+	run         = app.Command("run", "Runs the acceptance test suite")
+	runName     = run.Flag("name", "Name of Kubernetes context to against").Default("e2e").String()
+	contextName = "e2e"
 )
 
 // AcceptanceDockerfile defines the docker instructions used to create an acceptance
@@ -141,6 +142,7 @@ func main() {
 		logger.Log("msg", "successfully destroyed cluster")
 
 	case run.FullCommand():
+		contextName = *runName
 		RegisterFailHandler(Fail)
 
 		SetDefaultEventuallyTimeout(time.Minute)
@@ -154,11 +156,11 @@ func main() {
 	}
 }
 
-var _ = Specify("Acceptance", func() {
+var _ = Describe("Acceptance", func() {
 	ctx, cancel := signals.SetupSignalHandler()
 	defer cancel()
 
-	cfgPathBytes, err := exec.CommandContext(ctx, "kind", "get", "kubeconfig-path", "--name", *runName).Output()
+	cfgPathBytes, err := exec.CommandContext(ctx, "kind", "get", "kubeconfig-path", "--name", contextName).Output()
 	cfgPath := string(bytes.TrimSpace(cfgPathBytes))
 	if err != nil {
 		app.Fatalf("failed to discover kind cluster config path: %v", err)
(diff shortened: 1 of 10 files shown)


pkg/workloads/console/controller.go
```diff
@@ -166,15 +168,21 @@ func (r *reconciler) Reconcile() (res k8rec.Result, err error) {
...
+		// Requeue for when the console needs to be deleted
+		// In the future we could allow this to be configured in the template
+		res = requeueAfterInterval(r.logger, ConsoleTTL)
 	}
 
-	// TODO:
-	//   GC the terminated console if necessary
+	if r.console.Stopped() && time.Now().Sub(r.console.Status.ExpiryTime.Time) > ConsoleTTL {
+		r.logger.Log("event", EventDeleted, "resource", Console)
+		r.client.Delete(r.ctx, r.console)
```
benwh: I think you need to set a propagation policy on this, so that it cleans up the dependent resources?

pkg/workloads/console/controller.go
```diff
@@ -166,15 +168,21 @@ func (r *reconciler) Reconcile() (res k8rec.Result, err error) {
...
+	switch {
+	case r.console.Pending():
 		res = requeueAfterInterval(r.logger, time.Second)
-	case workloadsv1alpha1.ConsoleRunning:
+	case r.console.Running():
 		res = requeueForExpiration(r.logger, r.console.Status)
+	case r.console.Stopped():
+		// Requeue for when the console needs to be deleted
+		// In the future we could allow this to be configured in the template
+		res = requeueAfterInterval(r.logger, ConsoleTTL)
```
benwh: You may want to undo this in the conditional block below. Otherwise we're going to also try to reconcile this 24h *after* the console is deleted

pkg/workloads/console/controller.go
```diff
@@ -47,6 +47,8 @@ const (
 	ConsoleTemplate = "ConsoleTemplate"
 	Role            = "Role"
 	RoleBinding     = "RoleBinding"
+
+	ConsoleTTL = 24 * time.Hour
```
benwh: I wonder if instead of defining the TTL as a constant here we could set it in a func passed to `console.Add`? This could solve the test issue potentially

pkg/workloads/console/controller.go
hmac: True - although that's kinda harmless since we'll bail when we can't find the console.

pkg/workloads/console/controller.go
hmac: That's really sensible. Don't know why I didn't think of that 🤦‍♂️

pkg/workloads/console/controller.go
hmac: Good point!

pkg/workloads/console/controller.go
benwh: Yeah there's no harm, but best to avoid useless/confusing noise in the logs if we can I reckon

cmd/acceptance/main.go
```diff
@@ -141,6 +142,7 @@ func main() {
 		logger.Log("msg", "successfully destroyed cluster")
 
 	case run.FullCommand():
+		contextName = *runName
```
hmac: For some reason, locally, `*runName` is empty when the test below starts running. Explicitly tracking this in a normal string variable fixes this, but it's nasty and I'd like to remove it if possible.

pkg/apis/workloads/v1alpha1/console_types.go
```diff
@@ -93,6 +101,15 @@ type ConsoleTemplateSpec struct {
...
 	AdditionalAttachSubjects []rbacv1.Subject `json:"additionalAttachSubjects,omitempty"`
+
+	// Specifies the TTL for any Console created with this template. If set, the Console
+	// will be eligible for garbage collection ConsoleTTLSecondsAfterFinished seconds after
+	// it enters the Stopped phase. If not set, this value defaults to 24 hours.
+	// This field is modeled closely on the TTL mechanism in Kubernetes 1.12.
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=604800
+	TTLSecondsAfterFinished *int32 `json:"ttlSecondsAfterFinished,omitempty"`
```
hmac: I've gone for `*int32` here because it matches the type of the field of the same name on Jobs in Kube 1.12

PR #32, opened 2019-02-01
Add a console Runner package

benwh:
This is a wrapper around logic required to create and attach to
consoles, to be called from external clients.

The code here enables creation of a very basic console, and then waiting
for it to become ready to attach to.

The interfaces are not necessarily stable at this point and subject to change!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /bin
 .env
 .kube/
+*.coverprofile
--- a/pkg/workloads/console/runner/runner.go
+++ b/pkg/workloads/console/runner/runner.go
@@ -1,0 +1,161 @@
+package runner
+
+import (
+	"context"
+	"fmt"
+
+	workloadsv1alpha1 "github.com/gocardless/theatre/pkg/apis/workloads/v1alpha1"
+	"github.com/gocardless/theatre/pkg/client/clientset/versioned"
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Runner is responsible for managing the lifecycle of a console
+type Runner struct {
+	clientset versioned.Interface
+}
+
+// Options defines the parameters that can be set upon a new console
+type Options struct {
+	Cmd     []string
+	Timeout int
+
+	// TODO: For now we assume that all consoles are interactive, i.e. we setup a TTY on
+	// them when spawning them. This does not enforce a requirement to attach to the console
+	// though.
+	// Later on we may need to implement non-interactive consoles, for processes which
+	// expect a TTY to not be present?
+	// However with these types of consoles it will not be possible to send input to them
+	// when reattaching, e.g. attempting to send a SIGINT to cancel the running process.
+	// Interactive bool
+}
+
+// New builds a runner, with a given clientset
+func New(clientset versioned.Interface) *Runner {
+	return &Runner{
+		clientset: clientset,
+	}
+}
+
+// Create builds a console according to the supplied options and submits it to the API
+func (c *Runner) Create(namespace string, template workloadsv1alpha1.ConsoleTemplate, opts Options) (*workloadsv1alpha1.Console, error) {
+	csl := &workloadsv1alpha1.Console{
+		ObjectMeta: metav1.ObjectMeta{
+			// Let Kubernetes generate a unique name
+			GenerateName: template.Name + "-",
+
+			// TODO: Consider which labels and annotations we might want to set on the console
+			// itself.
+			// Labels:
+			// Annotations:
+		},
+		Spec: workloadsv1alpha1.ConsoleSpec{
+			ConsoleTemplateRef: corev1.LocalObjectReference{Name: template.Name},
+			// If the flag is not provided then the value will default to 0. The controller
+			// should detect this and apply the default timeout that is defined in the template.
+			TimeoutSeconds: opts.Timeout,
+			// TODO: This field is currently missing from the console spec
+			// Command:            c.Cmd,
+		},
+	}
+
+	return c.clientset.WorkloadsV1alpha1().Consoles(namespace).Create(csl)
+}
+
+// FindTemplateBySelector will search for a template matching the given label
+// selector and return errors if none or multiple are found (when the selector
+// is too broad)
+func (c *Runner) FindTemplateBySelector(namespace string, labelSelector string) (*workloadsv1alpha1.ConsoleTemplate, error) {
+	client := c.clientset.WorkloadsV1alpha1().ConsoleTemplates(namespace)
+
+	templates, err := client.List(
+		metav1.ListOptions{
+			LabelSelector: labelSelector,
+		},
+	)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to list consoles templates")
+	}
+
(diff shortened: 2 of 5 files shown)


pkg/workloads/console/runner/runner.go
```diff
@@ -0,0 +1,162 @@
+package runner
```
benwh: I'm not a big fan of this name. If anyone has any better suggestions then please shout!

pkg/workloads/console/runner/runner.go
```diff
@@ -0,0 +1,162 @@
...
+			// Labels:
+			// Annotations:
+		},
+		Spec: workloadsv1alpha1.ConsoleSpec{
+			ConsoleTemplateRef: corev1.LocalObjectReference{Name: template.Name},
+			// If the flag is not provided then the value will default to 0. The controller
+			// should detect this and apply the default timeout that is defined in the template.
+			TimeoutSeconds: opts.Timeout,
+			// TODO: This field is currently missing from the console spec
+			// Command:            c.Cmd,
```
benwh: I wanted this PR to stay tightly-scoped, so we can fix this in a follow-up

pkg/workloads/console/runner/runner.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	template := templates.Items[0]
+
+	return &template, nil
+}
+
+// WaitUntilReady will block until the console reaches a phase that indicates
+// that it's ready to be attached to, or has failed.
+func (c *Runner) WaitUntilReady(ctx context.Context, createdCsl workloadsv1alpha1.Console) error {
```
benwh: This whole function is a bit ugly. It should probably return machine-parseable errors, rather than error strings, and de-duplicate some of the state checking logic.

However I got a bit sick of writing tests for it at this point, and it does do the job we need for the moment, so would prefer to try and get it merged and continue work from there if possible!

.gitignore
```diff
@@ -1,3 +1,4 @@
 /bin
 .env
 .kube/
+*.coverprofile
```
benwh: These get generated in every directory when you run ginkgo / go test with `-cover`

pkg/workloads/console/runner/runner.go
```diff
@@ -0,0 +1,162 @@
...
+	isRunning := func(csl *workloadsv1alpha1.Console) bool {
+		return csl != nil && csl.Status.Phase == workloadsv1alpha1.ConsoleRunning
+	}
+	isStopped := func(csl *workloadsv1alpha1.Console) bool {
+		return csl != nil && csl.Status.Phase == workloadsv1alpha1.ConsoleStopped
+	}
```
hmac: There are helpers in https://github.com/gocardless/theatre/pull/31 that will help with this

pkg/workloads/console/runner/runner.go
hmac: I reckon this becomes a lot simpler if it returns `(*ConsolePhase, error)`, with the expectation it'll try to return `ConsoleRunning` if it can. Then you'd use it like this
```go
phase, err := runner.WaitUntilReady(..)
if err != nil {
 ...
}
if phase != workloadsv1alpha1.ConsoleRunning {
 ...
}
// now do your stuff
```

Then you only return errors if you genuinely can't fetch the console phase for some reason - and otherwise you just return the phase. Happy to leave this for a followup though.

pkg/workloads/console/runner/runner.go
benwh: I can see where you're coming from here, but I think that moving that method into anu would have a couple of downsides:
- It's going to be a lot harder/messier to test - I think that the tests are quitely nicely wrapped up at the moment
- Anyone else wanting to create consoles is going to have to duplicate the logic we've encoded in anu (while this is minimal at the moment, I expect that this method is going to grow a bit)

I did try and think about this for a while and played with the code in different places (after starting out with it all in anu) and this is the split that I disliked the least. It still doesn't feel totally natural though!

cc @lawrencejones, who probably has a more eloquent opinion on this

hmac: I'm confused - there's basically no logic in `Create`, right? We're just generating a name and then creating the console. What is there to test?

hmac: If we really want to leave `Create` in here, then it might be better as a convenience constructor, either in `types` or in a `consoles` package.

pkg/workloads/console/runner/runner.go
benwh: Yeah I think something like this is probably the way to go! I've made a note of it though and think that we should rework afterwards

pkg/workloads/console/runner/runner.go
benwh: Yep that'll help. Unfortunately we may still need the nil checks though

pkg/workloads/console/runner/runner.go
```diff
@@ -0,0 +1,162 @@
...
+
+	client := c.clientset.WorkloadsV1alpha1().Consoles(namespace)
+	return client.Create(csl)
+}
+
+// FindTemplateBySelector will search for a template matching the given label
+// selector and return errors if none or multiple are found (when the selector
+// is too broad)
+func (c *Runner) FindTemplateBySelector(namespace string, labelSelector string) (*workloadsv1alpha1.ConsoleTemplate, error) {
```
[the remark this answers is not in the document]
benwh: You mean as opposed to extracting the logic into the client side? It feels like enough code to not want to write more than once so I guess it's the same argument as the Create method, let's discuss up there!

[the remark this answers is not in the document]
hmac: Actually, I hadn't seen this was used in the client. This feels so common (find a resource by label) that surely there's a Kube function for it already?

pkg/workloads/console/runner/runner.go
benwh: > I'm confused - there's basically no logic in Create, right? We're just generating a name and then creating the console. What is there to test?

And also setting up the object reference to the template. And in the future copying over (some of?) the labels from the template to the console itself. We may also want to enforce that the console is created in the same namespace that the template exists in? It feels to me that it's a bit too early to tell how this'll play out.

Given that we're not doing any testing in anu, it feels nicer to be able to call an external interface such as this and have some confidence that it's not going to explode. But I accept that that argument doesn't hold too much weight in this case.

hmac: Fair, I buy that. Happy to go with this and we'll see if it grows into a horrible monster in time.

pkg/workloads/console/runner/runner.go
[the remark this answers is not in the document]
benwh: Not AFAIK, at least not in a generic way?

pkg/workloads/console/runner/runner.go
hmac: (really not a fan of the package name though. Could be called a processor or a broker or a manager or a handler or a supervisor and I'd just as much of a clue what it's meant to do)

pkg/workloads/console/runner/runner.go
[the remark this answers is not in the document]
hmac: ofc not - danke go

pkg/workloads/console/runner/runner.go
benwh: > Fair, I buy that. Happy to go with this and we'll see if it grows into a horrible monster in time.

Yeah I fully expect to land on the wrong side of history with this one! 😆

PR #33, opened 2019-02-01
Populate console job with command and set container properties

benwh:
If there's a non-empty command in the console spec, then use that rather
than whatever's in the template.

Also set the Stdin and TTY properties on the console container, to
ensure that it's possible to attach to it.

### DIFF
--- a/pkg/workloads/console/controller.go
+++ b/pkg/workloads/console/controller.go
@@ -395,6 +395,31 @@ func buildJob(name types.NamespacedName, csl *workloadsv1alpha1.Console, templat
 	timeout := int64(csl.Spec.TimeoutSeconds)
 
 	username := strings.SplitN(csl.Spec.User, "@", 2)[0]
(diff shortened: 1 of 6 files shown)


pkg/workloads/console/controller.go
```diff
@@ -395,14 +395,39 @@ func buildJob(name types.NamespacedName, csl *workloadsv1alpha1.Console, templat
...
+	numContainers := len(jobTemplate.Spec.Containers)
+
+	// If there's no containers in the spec then the controller will be emitting
+	// warnings anyway, as the job will be rejected
+	if numContainers > 0 {
+		c := jobTemplate.Spec.Containers[0]
+
+		// Only replace the template command if one is specified
+		if len(csl.Spec.Command) > 0 {
+			c.Command = csl.Spec.Command
```
benwh: I didn't create a test for the inverse of this condition, as it seemed overkill.
Maybe we can unit test this function in the future anyway, if we add more stuff to it.

PR #37, opened 2019-02-11
Use ginkgo test runner in CI

craigfurman:
It is customary, although not mandatory, to use the ginkgo test runner
when using the ginkgo test framework.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,6 +28,9 @@ jobs:
     steps:
       - checkout
       - run:
+          name: Install ginkgo test runner
+          command: go get github.com/onsi/ginkgo/ginkgo
+      - run:
           name: Install Kubebuilder test helpers
           command: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -35,7 +38,7 @@ jobs:
               | tar -xvz --strip=1 -C /usr/local/kubebuilder
       - run:
           name: Run tests
-          command: go test -race -count=1 -v ./...
+          command: ginkgo -race -r
```
benwh: Can we add the `-v` back in? I reckon it's always useful to have the extra info, because you're only going to look at the output if the tests fail

craigfurman: Can do! For the future, passing tests can be made silent and failing ones noisy by writing logs and other output to `ginkgo.GinkgoWriter`, which is an io.Writer that bins its buffer if the test passes, and spews it to stdout if it fails.

PR #41, opened 2019-02-12
Update example ConsoleTemplate metadata

craigfurman:
~~Add a namespace~~, and change labels to repo+env as currently used by the
only consumer to search for ConsoleTemplates.

### DIFF
--- a/config/samples/workloads_v1alpha1_consoletemplate.yaml
+++ b/config/samples/workloads_v1alpha1_consoletemplate.yaml
@@ -23,5 +23,5 @@ spec:
 metadata:
   name: console-template-0
   labels:
-    app: myapp
-    release: myapp-sandbox
+    repo: myapp-owner-myapp-repo
+    environment: myapp-env


craigfurman:
I suspect the CircleCI acceptance test step is broken, as this works for me* in a clean KinD cluster.

\*: 2019 equivalent of "it works on my machine"

config/samples/workloads_v1alpha1_consoletemplate.yaml
```diff
@@ -22,6 +22,7 @@ spec:
       restartPolicy: Never
 metadata:
   name: console-template-0
+  namespace: default
   labels:
-    app: myapp
-    release: myapp-sandbox
+    repo: myapp-owner-myapp-repo
+    env: myapp-env
```
lawrencejones: We've recently changed our label taxonomy to use `environment` everywhere. What are your thoughts on env vs environment here?

craigfurman: Not particularly fussed, will change

config/samples/workloads_v1alpha1_consoletemplate.yaml
```diff
@@ -22,6 +22,7 @@ spec:
       restartPolicy: Never
 metadata:
   name: console-template-0
+  namespace: default
```
benwh: Why add namespace here?

craigfurman: `namespace: default` is redundant, but in practice we will add namespaces to ConsoleTemplates, which the Console will inherit.

benwh: I still don't see that it adds any value in this example - this is obviously a namespaced resource so the object will have to go in *some* namespace. Adding this here just introduces the downside that if you were to pipe this through `kubectl apply` while targeting a different namespace in your current config, then the resource isn't going to go where you'd expect.

craigfurman: > if you were to pipe this through `kubectl apply` while targeting a different namespace in your current config, then the resource isn't going to go where you'd expect.

I was about to reply with "That isn't true" but it depends on what you mean by "expect"? I don't expect kubectl apply to only modify resources in my namespace, it depends on what's in the resource definitions.

In any case I'm not too fussed, I'll remove the namespace.

PR #43, opened 2019-02-13
Propagate labels onto console resources

craigfurman:
569d9ef: Propagate labels onto console resources

Copy user, repo, and environment labels from a console object to its
job, and the pod template in that job.

Also copy repo and environment labels from a ConsoleTemplate object to
Console objects in the `Create` factory function.

7004390: Console job and pod maintain user label info

Do not discard the domain of Console.Spec.User field values that
resemble email addresses.


craigfurman:
This is likely to close in favour of #44, after some discussion

PR #44, opened 2019-02-14
Enforce jobName and attach labels to job and pods

rnaveiras:
- JobName contains the postfix console and is truncated to ensure that is valid.
- Job has the same labels as console template and consoles. In the same way, a pod has the same labels as their job.

### DIFF
--- a/pkg/workloads/console/controller.go
+++ b/pkg/workloads/console/controller.go
@@ -12,6 +12,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 
@@ -423,11 +424,30 @@ func (r *reconciler) buildJob(template *workloadsv1alpha1.ConsoleTemplate) *batc
 	// 1 pod per job.
 	backoffLimit := int32(0)
 
+	// jobName: Add console postfix and truncate name no longer than 49
+	// characters (63-14) (14 characters '-console-4mh4z')
+	jobName := fmt.Sprintf("%s-%s", stringTruncate(r.name.Name, 49), "console")
+
+	// Merged labels from the console template and console
+	// In case of conflits second label set wins
+	jobLabels := labels.Merge(csl.Labels, template.Labels)
+	jobLabels = labels.Merge(jobLabels,
+		map[string]string{
+			"console-name": stringTruncate(csl.Name, 63),
+			"user":         sanitiseLabel(username),
+		})
+
+	// Add Pod lables
+	jobTemplate.ObjectMeta.Labels = labels.Merge(
+		jobLabels,
+		jobTemplate.ObjectMeta.Labels,
+	)
+
 	return &batchv1.Job{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      r.name.Name,
+			Name:      jobName,
 			Namespace: r.name.Namespace,
-			Labels:    map[string]string{"user": sanitiseLabel(username)},
+			Labels:    jobLabels,
 		},
 		Spec: batchv1.JobSpec{
 			Template:              *jobTemplate,
@@ -503,4 +523,11 @@ func jobDiff(expectedObj runtime.Object, existingObj runtime.Object) recutil.Out
 
 func sanitiseLabel(l string) string {
 	return regexp.MustCompile(`[^A-z0-9\-_.]`).ReplaceAllString(l, "-")
+}
+
+func stringTruncate(str string, length int) string {
+	if len(str) > length {
+		return str[0:length]
+	}
+	return str
 }
--- a/pkg/workloads/console/controller.go
+++ b/pkg/workloads/console/controller.go
@@ -424,20 +424,23 @@ func (r *reconciler) buildJob(template *workloadsv1alpha1.ConsoleTemplate) *batc
 	// 1 pod per job.
 	backoffLimit := int32(0)
 
-	// jobName: Add console postfix and truncate name no longer than 49
-	// characters (63-14) (14 characters '-console-4mh4z')
-	jobName := fmt.Sprintf("%s-%s", stringTruncate(r.name.Name, 49), "console")
+	// Ensure that the job name (after suffixing with `-console`) does not exceed 57
+	// characters, to allow an additional 6 characters to appended when the job
+	// creates a pod without truncation of the `-console` suffix.
+	jobName := fmt.Sprintf("%s-%s", truncateString(r.name.Name, 49), "console")
 
-	// Merged labels from the console template and console
-	// In case of conflits second label set wins
+	// Merged labels from the console template and console. In case of
+	// conflicts second label set wins.
+	// The labels on the console can be user-defined, so we do not want to allow a
+	// user to create a console with a label that implies that it's for an application
+	// different to the console.
 	jobLabels := labels.Merge(csl.Labels, template.Labels)
 	jobLabels = labels.Merge(jobLabels,
 		map[string]string{
-			"console-name": stringTruncate(csl.Name, 63),
+			"console-name": truncateString(csl.Name, 63),
 			"user":         sanitiseLabel(username),
 		})
 
-	// Add Pod lables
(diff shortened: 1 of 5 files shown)


rnaveiras:
Above shows jobName and propagate labels:

```
kubectl get consoletemplates,console,job,pod --show-labels
NAME                                                              AGE   LABELS
consoletemplate.workloads.crd.gocardless.com/console-template-0   36s   environment=myapp-env,repo=myapp-owner-myapp-repo

NAME                                             AGE   LABELS
console.workloads.crd.gocardless.com/console-0   7s    <none>

NAME                          COMPLETIONS   DURATION   AGE   LABELS
job.batch/console-0-console   0/1           7s         7s    console-name=console-0,environment=myapp-env,repo=myapp-owner-myapp-repo,user=kubernetes-admin

NAME                          READY   STATUS    RESTARTS   AGE   LABELS
pod/console-0-console-8bnh2   1/1     Running   0          7s    console-name=console-0,controller-uid=d35ab4a4-3069-11e9-b6b7-0242aa6d9636,environment=myapp-env,job-name=console-0-console,repo=myapp-owner-myapp-repo,user=kubernetes-admin
```

pkg/workloads/console/controller.go
```diff
@@ -423,11 +424,30 @@ func (r *reconciler) buildJob(template *workloadsv1alpha1.ConsoleTemplate) *batc
 	// 1 pod per job.
 	backoffLimit := int32(0)
 
+	// jobName: Add console postfix and truncate name no longer than 49
+	// characters (63-14) (14 characters '-console-4mh4z')
+	jobName := fmt.Sprintf("%s-%s", stringTruncate(name.Name, 49), "console")
```
benwh: `name` no longer exists, so this needs to be prefixed with `r.`

pkg/workloads/console/controller.go
```diff
@@ -423,11 +424,30 @@ func (r *reconciler) buildJob(template *workloadsv1alpha1.ConsoleTemplate) *batc
 	// 1 pod per job.
 	backoffLimit := int32(0)
 
+	// jobName: Add console postfix and truncate name no longer than 49
+	// characters (63-14) (14 characters '-console-4mh4z')
```
benwh: Entirely subjective, but, this comment might be clearer if it said something like:
> Ensure that the job name (after suffixing with `-console`) does not exceed 57 characters, to allow an additional 6 characters to appended when the job creates a pod without truncation of the `-console` suffix.

pkg/workloads/console/controller.go
```diff
@@ -423,11 +424,30 @@ func (r *reconciler) buildJob(template *workloadsv1alpha1.ConsoleTemplate) *batc
...
+	jobName := fmt.Sprintf("%s-%s", stringTruncate(r.name.Name, 49), "console")
+
+	// Merged labels from the console template and console
+	// In case of conflits second label set wins
+	jobLabels := labels.Merge(csl.Labels, template.Labels)
+	jobLabels = labels.Merge(jobLabels,
+		map[string]string{
+			"console-name": stringTruncate(csl.Name, 63),
```
benwh: Minor: This method should probably be called `truncateString`, for consistency with `sanitiseLabel`

pkg/workloads/console/controller.go
```diff
@@ -423,11 +424,30 @@ func (r *reconciler) buildJob(template *workloadsv1alpha1.ConsoleTemplate) *batc
 	// 1 pod per job.
 	backoffLimit := int32(0)
 
+	// jobName: Add console postfix and truncate name no longer than 49
+	// characters (63-14) (14 characters '-console-4mh4z')
+	jobName := fmt.Sprintf("%s-%s", stringTruncate(r.name.Name, 49), "console")
+
+	// Merged labels from the console template and console
+	// In case of conflits second label set wins
```
benwh: Typo: `conflicts`

I think that this comment might be better if we explained *why* we're giving the labels from the template precedence; i.e.
> The labels on the console can be user-defined, so we do not want to allow a user to create a console with a label that implies that it's for an application different to the console.

pkg/workloads/console/controller.go
```diff
@@ -423,11 +424,30 @@ func (r *reconciler) buildJob(template *workloadsv1alpha1.ConsoleTemplate) *batc
...
+	jobLabels := labels.Merge(csl.Labels, template.Labels)
+	jobLabels = labels.Merge(jobLabels,
+		map[string]string{
+			"console-name": stringTruncate(csl.Name, 63),
+			"user":         sanitiseLabel(username),
+		})
+
+	// Add Pod lables
```
benwh: I think we can kill this comment (and the typo with it 🙈)

pkg/workloads/console/integration/integration_test.go
```diff
@@ -188,15 +192,38 @@ var _ = Describe("Console", func() {
...
 			job := &batchv1.Job{}
 			identifier, _ := client.ObjectKeyFromObject(csl)
+			identifier.Name += "-console"
 			err := mgr.GetClient().Get(context.TODO(), identifier, job)
 
 			Expect(err).NotTo(HaveOccurred(), "failed to find associated Job for Console")
+
+			By("Expect job and pod labels are correctly populated from the console template and console")
+			Expect(
+				job.ObjectMeta.Labels["user"]).To(Equal("system-unsecured"))
```
benwh: Missing a description parameter on the `To` function (to tell us the issue if this assertion fails)
e.g.
> job should have a user label matching the console owner

Applies to the expectation below also

pkg/workloads/console/integration/integration_test.go
```diff
@@ -188,15 +192,38 @@ var _ = Describe("Console", func() {
...
+			Expect(
+				job.ObjectMeta.Labels["console-name"]).To(Equal(csl.ObjectMeta.Name))
+			Expect(
+				job.ObjectMeta.Labels["repo"]).To(Equal(consoleTemplate.ObjectMeta.Labels["repo"]),
+				"job's has same labels as the console template (preference over console labels)",
```
benwh: grammar:
> job has the same labels ...

(repeated below)

pkg/workloads/console/integration/integration_test.go
```diff
@@ -188,15 +192,38 @@ var _ = Describe("Console", func() {
...
+			)
+			Expect(
+				job.ObjectMeta.Labels["other-label"]).To(Equal(csl.ObjectMeta.Labels["other-label"]),
+				"job's has same labels as the console",
+			)
+			Expect(
+				job.Spec.Template.Labels["job-name"]).To(Equal(job.ObjectMeta.Name),
```
benwh: I'm a bit confused where this is coming from - is the apiserver automatically setting it when we create the job object?

If that's the case, then we probably shouldn't be testing it here?

pkg/workloads/console/integration/integration_test.go
```diff
@@ -312,13 +339,37 @@ var _ = Describe("Console", func() {
 			Expect(csl.ObjectMeta.OwnerReferences[0].Name).To(Equal(consoleTemplate.ObjectMeta.Name))
 		})
 	})
+
+	Describe("Enforcing job name", func() {
+		It("Job names are less than 49 characters and contains postfix 'console'", func() {
```
benwh: `It("Truncates long job names and adds a 'console' suffix")` reads better?

pkg/workloads/console/runner/runner_test.go
```diff
@@ -74,6 +77,10 @@ var _ = Describe("Runner", func() {
 				Expect(list.Items).To(HaveLen(1), "only one console should be present")
 			})
 
+			It("Inheritance labels from console template", func() {
```
benwh: > Inherits labels ...

pkg/workloads/console/runner/runner_test.go
```diff
@@ -366,23 +361,18 @@ var _ = Describe("Runner", func() {
...
-					Labels:    map[string]string{"job-name": job.ObjectMeta.Name},
-				},
-				Spec: corev1.PodSpec{
-					Containers: []corev1.Container{corev1.Container{TTY: true}},
-				},
-			}
 			BeforeEach(func() {
-				fakeKubeObjects = []runtime.Object{job, pod2}
+				consolePod.Spec = corev1.PodSpec{
+					Containers: []corev1.Container{corev1.Container{TTY: true}},
```
benwh: The second `corev1.Container` can be omitted

pkg/workloads/console/integration/integration_test.go
rnaveiras: I agreed that now it doesn't make have this here.

PR #46, opened 2019-02-14
Add events tests

benwh:
This PR is very much WIP.

We want to assert on the correct events being emitted in response to reconciliation of different objects.

However the current implementation in this branch is flaky to the point of being unusable, because the default events sink is only guaranteed to flush once every 10 seconds.

There's a few of potential avenues to work around this:
1. Inject something like `record.FakeRecorder` (from `client-go`)
1. Assert on the output of the controller output (which would involve parsing the logfmt lines)
1. Make the logging sink flush synchronously


lawrencejones:
Closing as stale, can re-open if we need it.

PR #48, opened 2019-02-14
Replace RoleBinding for DirectoryRoleBinding

rnaveiras:
Support DirectoryRoleBinding for console additionalAttachSubjects

### DIFF
--- a/config/managers/workloads.yaml
+++ b/config/managers/workloads.yaml
@@ -28,7 +28,6 @@ rules:
       - rbac.authorization.k8s.io
     resources:
       - roles
-      - rolebindings
     verbs:
       - "*"
   - apiGroups:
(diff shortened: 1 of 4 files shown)


pkg/recutil/reconcile.go
```diff
@@ -176,3 +178,23 @@ func RoleBindingDiff(expectedObj runtime.Object, existingObj runtime.Object) Out
 
 	return operation
 }
+
```
benwh: We could now remove the `RoleBindingDiff` func. I don't think there's any use for it; as the RBAC controller does things in a different way

config/managers/workloads.yaml
```diff
@@ -41,6 +41,12 @@ rules:
       - validatingwebhookconfigurations
     verbs:
       - "*"
+  - apiGroups:
```
benwh: Should we now remove the `rbac.authorization.k8s.io/rolebindings` permissions?

pkg/recutil/reconcile.go
rnaveiras: I removed it, it's no need it anymore.

config/managers/workloads.yaml
rnaveiras: Done

PR #49, opened 2019-02-15
Create Role with the console Pod as resourceNames

rnaveiras:
Create the Role and the DirectoryRoleBinding after the Pod Phase
transition to Pod Running. It's required to know the Pod Name, therefore
the role created target only that specific Pod resource.

Also updates the role with the required permissions like pod/exec,
pod/attach and delete pod.

### DIFF
--- a/config/managers/workloads.yaml
+++ b/config/managers/workloads.yaml
@@ -46,6 +46,28 @@ rules:
       - directoryrolebindings
     verbs:
       - "*"
+  # The following permissions are provided to allow the manager to create roles
+  # with these permissions
+  - apiGroups:
+      - ""
+    resources:
+      - "pods"
+    verbs:
+      - get
+      - delete
+  - apiGroups:
+      - ""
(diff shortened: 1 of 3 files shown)


rnaveiras:
Example of role created after the pod is running:
```
$ k describe role
Name:         console-0
Labels:       <none>
Annotations:  <none>
PolicyRule:
  Resources         Non-Resource URLs  Resource Names             Verbs
  ---------         -----------------  --------------             -----
  pods.core/attach  []                 [console-0-console-zdgbn]  [create]
  pods.core/exec    []                 [console-0-console-zdgbn]  [create]
  pods.core         []                 [console-0-console-zdgbn]  [get delete]
```

lawrencejones:
Let's add pod logs here as well whilst we're touching. I know we plan to add log access for everyone anyway, but there should be no case where you can exec and attach to a pod but not view it's logs.

PR #54, opened 2019-02-19
Always set a console pod's RestartPolicy to Never

benwh:
In order to more easily reason about the lifetime of our console tasks,
we never want a container that fails with a non-zero exit code to [be
restarted][1].

The only other supported value for `RestartPolicy` in Jobs is
`OnFailure`, which we won't want to use, so setting it to `Never` in the
controller is a sensible step.

This saves us from having to specify a RestartPolicy in each template
that we define.

[1]: https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#handling-pod-and-container-failures

### DIFF
--- a/pkg/workloads/console/integration/integration_test.go
+++ b/pkg/workloads/console/integration/integration_test.go
@@ -243,6 +243,10 @@ var _ = Describe("Console", func() {
 				"job's BackoffLimit is not 0",
 			)
 			Expect(
+				job.Spec.Template.Spec.RestartPolicy).To(Equal(corev1.RestartPolicyNever),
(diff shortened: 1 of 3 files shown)


pkg/workloads/console/integration/integration_test.go
```diff
@@ -435,7 +439,7 @@ func buildConsoleTemplate(namespace string) *workloadsv1alpha1.ConsoleTemplate {
 							Command: []string{"/bin/sh", "-c", "sleep 100"},
 						},
 					},
-					RestartPolicy: "Never",
+					RestartPolicy: "OnFailure",
```
craigfurman: 👌 Good regression test

PR #57, opened 2019-02-20
WaitUntilReady waits for RoleBinding

craigfurman:
The Console controller provisions a DirectoryRoleBinding to bind the
console user (and any additional attach subjects) to a role that grants
them permission to attach to the console pod. The DirectoryRoleBinding
controller then provisions a RoleBinding, and then updates it with the
subject list.

Until this RoleBinding exists, and contains the user in its subject
list, the user will not be able to use their console.

### DIFF
--- a/pkg/workloads/console/runner/runner.go
+++ b/pkg/workloads/console/runner/runner.go
@@ -8,6 +8,7 @@ import (
 	"github.com/gocardless/theatre/pkg/client/clientset/versioned"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -149,7 +150,23 @@ func (c *Runner) ListConsolesByLabelsAndUser(namespace, username, labelSelector 
 
 // WaitUntilReady will block until the console reaches a phase that indicates
 // that it's ready to be attached to, or has failed.
+// It will then block until an associated RoleBinding exists that contains the
+// console user in its subject list. This RoleBinding gives the console user
+// permission to attach to the pod.
 func (c *Runner) WaitUntilReady(ctx context.Context, createdCsl workloadsv1alpha1.Console) (*workloadsv1alpha1.Console, error) {
+	csl, err := c.waitForConsole(ctx, createdCsl)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := c.waitForRoleBinding(ctx, csl); err != nil {
+		return nil, err
+	}
+
+	return csl, nil
+}
+
+func (c *Runner) waitForConsole(ctx context.Context, createdCsl workloadsv1alpha1.Console) (*workloadsv1alpha1.Console, error) {
 	isRunning := func(csl *workloadsv1alpha1.Console) bool {
 		return csl != nil && csl.Status.Phase == workloadsv1alpha1.ConsoleRunning
 	}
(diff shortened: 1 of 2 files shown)


pkg/workloads/console/runner/runner.go
```diff
@@ -211,6 +228,56 @@ func (c *Runner) WaitUntilReady(ctx context.Context, createdCsl workloadsv1alpha
 	}
 }
 
+func (c *Runner) waitForRoleBinding(ctx context.Context, csl *workloadsv1alpha1.Console) error {
```
lawrencejones: This is quite tightly coupled to our implementation of our controller. Do you think it might be better if we use a `SelfSubjectAccessReview` to instead ask the cluster if I have permissions to attach? That's asking a more direct question of whether we're ready and able to continue.

pkg/workloads/console/runner/runner_test.go
```diff
@@ -282,125 +287,275 @@ var _ = Describe("Runner", func() {
...
+
+			readyRb = rbacv1.RoleBinding{
+				ObjectMeta: metav1.ObjectMeta{Name: csl.Name},
+				Subjects:   []rbacv1.Subject{{Name: csl.Spec.User}},
+			}
+		})
+
+		Describe("waiting for the console to be ready", func() {
 
 			BeforeEach(func() {
```
lawrencejones: This style - the space after the describe - is totally novel afaik? Can we cut it back so this file looks like the rest of our tests?

pkg/workloads/console/runner/runner.go
```diff
@@ -211,6 +228,56 @@ func (c *Runner) WaitUntilReady(ctx context.Context, createdCsl workloadsv1alpha
...
+	rbClient := c.kubeClient.RbacV1().RoleBindings(csl.Namespace)
```
lawrencejones: Does this become much simpler both to implement and test if we poll?

```go
for {
	select {
	case <-ctx.Done():
		return errors.New("timed out waiting for rolebinding to be created")
	default:
		if rb, err := rbClient.Get(csl.Name, metav1.GetOptions{}); err == nil {
			break // found
		}

		time.Sleep(interval)
	}
}
```

pkg/workloads/console/runner/runner_test.go
craigfurman: I thought it was weird but it's everywhere in the theatre test suites! I thought it was you, and I was trying to blend in. Will cut these lines.

pkg/workloads/console/runner/runner.go
craigfurman: I'll look into this, I didn't know about SelfSubjectAccessReview. I agree it's a more blackbox way of verifying behaviour!

pkg/workloads/console/runner/runner.go
craigfurman: It does, but the using a watcher felt like a more "kube-native" (™️) way of doing this, and there was already analogous code right next to this using a console watcher. You still need to perform a Get to mitigate the race with opening the watcher. I'll remove all watcher code from both polling loops and we can see how that looks.

pkg/workloads/console/runner/runner_test.go
lawrencejones: Merci beaucoup, it freaks me out

pkg/workloads/console/runner/runner.go
lawrencejones: Yeah I hadn't properly looked at the rest of the file. I now see we're repeating this pattern everywhere.

Two things:

- I think this is the proper way of doing this
- I think the way we do it right now complicates our code

The second part wouldn't be half as bad if it didn't commit us to making any stubbing tests quite nasty, and introducing lots of async pain to our testing setup. We can solve this by either using more realistic tests (really interacting with kubernetes) or by creating a better concept of "watch this thing, but also inject the current value if it's already there".

Given some tests will always be unit tests, in the past I've gone with creating an abstraction I mentioned above, the "get and watch" or however you term it. You then properly test that code and can trust the implementation to do the right thing elsewhere, and for unit tests replace it with a reliable mock.

This is great and all, but I didn't realise I was opening pandoras box here. I think we might be better off going with this implementation now if it works and then ticketing this up as a possible improvement later. Perhaps even just opening an issue in the github repo about this as an optimisation. I'm keen to merge something that works for this now to enable console rollout and revisit the implementation later.

Wdyt?

craigfurman: Totally agree! Let's merge this now as it works and I'll PR that improvement next. Small steps and all.

pkg/workloads/console/runner/runner_test.go
```diff
@@ -282,30 +281,88 @@ var _ = Describe("Runner", func() {
...
+				Spec:       workloadsv1alpha1.ConsoleSpec{User: "test-user"},
+			}
+
+			readyRb = rbacv1.RoleBinding{
+				ObjectMeta: metav1.ObjectMeta{Name: csl.Name},
+				Subjects:   []rbacv1.Subject{{Name: csl.Spec.User}},
+			}
+		})
 
+		Describe("waiting for the console to be ready", func() {
```
lawrencejones: We capitalise all the Describe/Context/Its

pkg/workloads/console/runner/runner.go
craigfurman: Similar to thread below, let's merge this now because it works, and I'll PR this improvement v soon.

PR #59, opened 2019-02-22
WIP: CircleCI caches GOCACHE

craigfurman:
Work in progress.


lawrencejones:
We'll want to be switching to go mod, at which point this will become unnecessary.

PR #64, opened 2019-04-26
Rename platform@ to core-infrastructure-team@

nlopes:


lawrencejones:
I don't think we should change this as these are all test fixtures and don't need to be wrong. Kinda besides the point now but we probably also don't want to publicise this email :see_no_evil:

PR #65, opened 2019-06-11
WIP: Support deleting consoles by name

hmac:


benwh:
Closing this as stale. If we do want this feature in the future then we can potentially keep some of this code, but also would want to build it into the `theatre-consoles` CLI

PR #71, opened 2019-10-21
DirectoryRoleBinding: fix double reconcile loop

lawrencejones:
Ensure our directoryrolebinding controller annotates the error with the
intended result, so we can better understand what is failing when we see
errors in the logs.

### DIFF
--- a/pkg/rbac/directoryrolebinding/controller.go
+++ b/pkg/rbac/directoryrolebinding/controller.go
@@ -104,7 +104,7 @@ func (r *Reconciler) ReconcileObject(logger kitlog.Logger, request reconcile.Req
 			return reconcile.Result{}, errors.Wrap(err, "failed to get DirectoryRoleBinding")
 		}
 
-		rb := &rbacv1.RoleBinding{
+		rb = &rbacv1.RoleBinding{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      drb.Name,
 				Namespace: drb.Namespace,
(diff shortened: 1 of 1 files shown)


pkg/rbac/directoryrolebinding/controller.go
```diff
@@ -99,11 +100,11 @@ func (r *Reconciler) ReconcileObject(logger kitlog.Logger, request reconcile.Req
...
 	err = r.client.Get(r.ctx, identifier, rb)
 	if err != nil {
-		if !errors.IsNotFound(err) {
-			return reconcile.Result{}, err
+		if !apierrors.IsNotFound(err) {
+			return reconcile.Result{}, errors.Wrap(err, "failed to get DirectoryRoleBinding")
 		}
 
-		rb := &rbacv1.RoleBinding{
+		rb = &rbacv1.RoleBinding{
```
dyson: Oh wow this is so easy to do isn't it!

pkg/rbac/directoryrolebinding/controller.go
```diff
@@ -6,9 +6,10 @@ import (
 	"time"
 
 	kitlog "github.com/go-kit/kit/log"
+	"github.com/pkg/errors"
 
 	rbacv1 "k8s.io/api/rbac/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
```
dyson: I thought this would have been an error importing two packages with the same name. Interesting.

lawrencejones: Why would it? You've aliased the packages to avoid that, right?

dyson: Oh no my bad. I thought 'pkg/errors' was already in there 🤦‍♂so was wondering how it ever worked before you aliased it.

PR #72, opened 2019-11-11
Add theatre-envconsul

dyson:
This PR adds a binary to be used by a muting webhook (https://github.com/gocardless/theatre/pull/73). 

The install command copies the theatre-envconsul and envconsul binaries
from the root filesystem (where they are expected to be installed by our
docker image) to an install path. We intend to run this as an init
container.

The exec command then authenticates with Vault, exchanging a kubernetes
service account token (detected from our environment) for a Vault token.
We then parse user configuration files which contain an environment
variable map, and produce an envconsul file that will resolve vault
secrets for those environment variables that have a `vault:` prefix. We
finally os.Exec to the envconsul command, which then takes over.

We expect the exec command to be prepended to pod container commands.

Specifying vault secrets via both environment variables and/or a configuration file is supported. The  configuration file takes precedence over an environment variable of the same name.

The configuration file is yaml and reads the env key for key/value pairs. Values that are prefixed with `vault:` are assumed to be a vault path to a secret, everything else is passed through as a normal environment variable. Under the hood we configure envconsul to fetch the secret and set the environment variable to be the value of it, replacing the path.
```
---
  environment:
    ENV_VAR_NAME: "not a secret"
    SECRET_ENV_VAR_NAME: "vault:/path/to/secret/"
```

The same rule is applied for environment variables. If the pod has an environment variable that has `vault:` as the prefix of its value, it will be replaced with the contents of the secret at that path. 

```
export SECRET_ENV_VAR_NAME_TWO=vault:/path/to/secret-two/
```


Dockerfile
```diff
@@ -2,19 +2,33 @@
...
+                            unzip \
       && apt-get clean -y \
       && rm -rf /var/lib/apt/lists/*
 
+# Install envconsul for theatre-envconsul
+ENV ENVCONSUL_VERSION=0.9.1 \
+    ENVCONSUL_SHA256=b58d032ad61937eca9def17482807124fa1bafac7e7bb5e025ea8a28d9c6ce42
+
+RUN set -x \
+  && curl -o /tmp/envconsul.zip -fsL https://releases.hashicorp.com/envconsul/0.9.1/envconsul_${ENVCONSUL_VERSION}_linux_amd64.zip \
```
lawrencejones: Add an extra tab to align with the above

pkg/file/copy.go
```diff
@@ -0,0 +1,55 @@
+package file
```
lawrencejones: I wouldn't have a package for this- I'd put this as a function inside the `cmd/theatre-envconsul/main.go` personally. It won't be used anywhere other than that!

On an aside, christ this is totally horrible. Why is this not simpler.

pkg/file/copy.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/pkg/errors"
+)
+
+func Copy(src, dst string, bufferSize int64) error {
```
lawrencejones: Assuming that we'll only ever be copying a standard file, can we implement this more simply as:

```go
srcFile, err := os.Open(src)
if err != nil {
  return err
}

defer srcFile.Close()

dstFile, err := os.Create(dst)
if err != nil {
  return err
}

defer dstFile.Close()

if _, err := io.Copy(dstFile, srcFile); err != nil {
  return err
}
```

?

lawrencejones:
```ruby
FileUtils.cp(src, dst)
```

:trollface:

dyson: I feel yuck copying a file fully into memory! But yeah, we know these are small, and as you point out, regular files, so let's simplify it.

pkg/file/copy.go
dyson: I guess everyone has different needs from a copy command? And then handling all the edge cases is just verbose. But like you mention below, we know what files we are copying, so we can get away with assuming things.

Dockerfile
dyson: Good spot!

pkg/file/copy.go
lawrencejones: That doesn't copy everything into memory? `io.Copy` buffers internally

dyson: Yeah 32mb. I guess I'm old and remember the days when I had less ram than that in my whole PC!

lawrencejones: Dude, you're copying it into an on-memory filesystem :joy:

lawrencejones:
@rnaveiras / @benwh would be great to get your feedback on this, as @dyson @mos3abof and I have all worked on it at some point.

cmd/theatre-envconsul/main.go
```diff
@@ -0,0 +1,314 @@
...
+
+		logger.Log("msg", "copying files into install path", "file_path", *installPath)
+		for _, f := range files {
+			if err := copyExecutable(f, path.Join(*installPath, f)); err != nil {
+				return errors.Wrap(err, "error copying file")
+			}
+		}
+
+	case exec.FullCommand():
+		clusterConfig, err := rest.InClusterConfig()
```
benwh: I think it' helpful to support running this outside of Kubernetes too. This gives a much quicker feedback loop, and has been what I've been using for testing.

The changes to allow this are pretty simple, e.g. (most of this is just whitespace changes)
```diff
diff --git cmd/theatre-envconsul/main.go cmd/theatre-envconsul/main.go
index 3ecd055..676dc91 100644
--- cmd/theatre-envconsul/main.go
+++ cmd/theatre-envconsul/main.go
@@ -46,6 +46,7 @@ var (
 	execCommand              = exec.Flag("command", "Command to execute").Required().String()
 	execInstallPath          = exec.Flag("install-path", "Path containing installed binaries").Default(defaultInstallPath).String()
 	execVaultAddress         = exec.Flag("vault-address", "Address of vault (format: scheme://host:port)").Required().String()
+	execVaultToken           = exec.Flag("vault-token", "Vault token to use, instead of Kubernetes auth").OverrideDefaultFromEnvar("VAULT_TOKEN").String()
 
 	// Version is set at compile time
 	Version = "dev"
@@ -81,15 +82,21 @@ func mainError(ctx context.Context, command string) (err error) {
 		}
 
 	case exec.FullCommand():
-		clusterConfig, err := rest.InClusterConfig()
-		if err != nil {
-			return errors.Wrap(err, "failed to authenticate within kubernetes")
-		}
+		var vaultToken string
+
+		if *execVaultToken != "" {
+			vaultToken = *execVaultToken
+		} else {
+			clusterConfig, err := rest.InClusterConfig()
+			if err != nil {
+				return errors.Wrap(err, "failed to authenticate within kubernetes")
+			}
 
-		logger.Log("event", "vault_login", "host", *execVaultAddress, "backend", *execAuthBackendMountPath, "role", *execAuthBackendRole)
-		vaultToken, err := getVaultToken(clusterConfig.BearerToken, *execVaultAddress, *execAuthBackendMountPath, *execAuthBackendRole)
-		if err != nil {
-			return errors.Wrap(err, "failed to login to vault")
+			logger.Log("event", "vault_login", "host", *execVaultAddress, "backend", *execAuthBackendMountPath, "role", *execAuthBackendRole)
+			vaultToken, err = getVaultToken(clusterConfig.BearerToken, *execVaultAddress, *execAuthBackendMountPath, *execAuthBackendRole)
+			if err != nil {
+				return errors.Wrap(err, "failed to login to vault")
+			}
 		}
 
 		logger.Log("event", "loading_config", "file_path", *execConfigFile)

```

cmd/theatre-envconsul/main.go
```diff
@@ -0,0 +1,314 @@
...
+	defer srcFile.Close()
+
+	dstFile, err := os.Create(dst)
+	if err != nil {
+		return err
+	}
+	defer dstFile.Close()
+
+	if _, err := io.Copy(dstFile, srcFile); err != nil {
+		return errors.Wrapf(err, "error copying %s -> %s", src, dst)
```
benwh: nit: Error message should be `failed to copy ...`, for consistency

cmd/theatre-envconsul/main.go
```diff
@@ -68,34 +69,31 @@ func main() {
...
-			return errors.Wrap(err, "failed to authenticate within kubernetes")
-		}
-
-		logger.Log("event", "vault_login", "host", *execVaultAddress, "backend", *execAuthBackendMountPath, "role", *execAuthBackendRole)
-		vaultToken, err := getVaultToken(clusterConfig.BearerToken, *execVaultAddress, *execAuthBackendMountPath, *execAuthBackendRole)
-		if err != nil {
-			return errors.Wrap(err, "failed to login to vault")
-		}
+		var vaultToken string
+		if *execVaultToken != "" {
```
jace-ys: this should be `==` i think. i could only get a token after providing a junk string to `--vault-token` 😆

lawrencejones: Yep, me being silly. Thanks!

cmd/theatre-envconsul/main.go
lawrencejones: What token were you using for testing Ben? Your Vault account shouldn't permit retrieval of secrets, so it's a bit worrying if you were able to pull them!

benwh: I was using a local vault server in dev mode

lawrencejones: Ah gotcha, thought this was lab for a minute

lawrencejones:
Let's get the first version of this into master, so we can make smaller incremental changes after.

PR #73, opened 2019-11-19
Add envconsul mutating webhook

dyson:
This PR introduces a mutating webhook for injecting application secrets from vault as env vars. It does this by creating an init container to inject `theatre-envconsul` (https://github.com/gocardless/theatre/pull/72) and `envconsul` into the applications pod, as well as replacing the pods command with the `theatre-envconsul exec` shim. 

Unfortunately we can't specify which pods to mutate based on annotations so we register the webhook for all pods in all namespaces. In the webhook we check for the presence of the correct annotation and if there isn't one, we return the pod without mutation. If the pods contains the correct annotations, we will mutate it. 

The annotation is namespaced to `envconsul-injector.vault.crd.gocardless.com` and we read the `configs` key. The value of the annotation is a comma separated list of container names that we should mutate. Optionally the file path to a configuration file can also be specified after the container name.

Example annotation for a container with a configuration file:
```
annotations:
    envconsul-injector.vault.crd.gocardless.com/configs: "app:/run/vault/app-config.yaml"
```

Example annotation for a container without a configuration file:
```
annotations:
    envconsul-injector.vault.crd.gocardless.com/configs: "app"
```

Example of a mix of containers with and without configuration files:
```
annotations:
    envconsul-injector.vault.crd.gocardless.com/configs: "app1,app2:/run/vault/app2-config.yaml,app3"
```

The format of the configuration file is given in the PR for theatre-envconsul (https://github.com/gocardless/theatre/pull/72).

### DIFF
--- a/cmd/vault-manager/main.go
+++ b/cmd/vault-manager/main.go
@@ -1,0 +1,96 @@
+package main
+
+import (
+	"os"
+
+	"github.com/alecthomas/kingpin"
+
+	"k8s.io/client-go/kubernetes/scheme"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp" // this is required to auth against GCP
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/gocardless/theatre/cmd"
+	"github.com/gocardless/theatre/pkg/apis"
+	"github.com/gocardless/theatre/pkg/signals"
+	"github.com/gocardless/theatre/pkg/vault/envconsul"
+)
+
+var (
+	app                     = kingpin.New("vault-manager", "Manages vault.crd.gocardless.com resources").Version(Version)
+	namespace               = app.Flag("namespace", "Kubernetes webhook service namespace").Default("theatre-system").String()
+	serviceName             = app.Flag("service-name", "Webhook service name").Default("theatre-vault-manager").String()
+	webhookName             = app.Flag("webhook-name", "Mutating webhook name").Default("theatre-vault").String()
+	theatreImage            = app.Flag("theatre-image", "Set to the same image as current binary").Required().String()
+	installPath             = app.Flag("install-path", "Location to install theatre binaries").Default("/var/run/theatre").String()
+	vaultConfigMapName      = app.Flag("vault-configmap-name", "Vault configMap name containing vault configuration").Default("vault-config").String()
+	vaultConfigMapNamespace = app.Flag("vault-configmap-namespace", "Namespace of vault configMap").Default("vault-system").String()
+
+	commonOpts = cmd.NewCommonOptions(app)
+
+	// Version is set at compile time
+	Version = "dev"
+)
+
+func main() {
+	kingpin.MustParse(app.Parse(os.Args[1:]))
+	logger := commonOpts.Logger()
+
+	if err := apis.AddToScheme(scheme.Scheme); err != nil {
+		app.Fatalf("failed to add schemes: %v", err)
+	}
+
+	ctx, cancel := signals.SetupSignalHandler()
+	defer cancel()
+
+	mgr, err := manager.New(config.GetConfigOrDie(), manager.Options{})
+	if err != nil {
+		app.Fatalf("failed to create manager: %v", err)
+	}
+
+	opts := webhook.ServerOptions{
+		CertDir: "/tmp/theatre-vault",
+		BootstrapOptions: &webhook.BootstrapOptions{
+			MutatingWebhookConfigName: *webhookName,
+			Service: &webhook.Service{
+				Namespace: *namespace,
+				Name:      *serviceName,
+				Selectors: map[string]string{
+					"app":   "theatre",
+					"group": "vault.crd.gocardless.com",
+				},
+			},
+		},
+	}
+
+	svr, err := webhook.NewServer("vault", mgr, opts)
+	if err != nil {
+		app.Fatalf("failed to create admission server: %v", err)
+	}
+
+	injectorOpts := envconsul.InjectorOptions{
+		Image:       *theatreImage,
+		InstallPath: *installPath,
+		VaultConfigMapKey: client.ObjectKey{
+			Namespace: *vaultConfigMapNamespace,
+			Name:      *vaultConfigMapName,
+		},
+	}
+
+	var wh *admission.Webhook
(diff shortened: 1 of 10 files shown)


pkg/vault/envconsul/webhook.go
```diff
@@ -0,0 +1,177 @@
...
+	vaultAuthMountPath, _ := vaultConfig.Data["auth_mount_path"]
+	vaultAuthRole, _ := vaultConfig.Data["auth_role"] // TODO: Add to the config file
+
+	webhookConfig := &corev1.ConfigMap{}
+	i.client.Get(ctx, i.webhookConfigKey, webhookConfig)
+	image, _ := webhookConfig.Data["image"]
+	installPath, _ := webhookConfig.Data["install_path"]
+
+	containerConfigPaths := getContainerConfigPaths(config)
+	copy := pod.DeepCopy()
```
benwh: This variable name feels a bit weird. How about `newPod`?

pkg/vault/envconsul/webhook.go
```diff
@@ -0,0 +1,177 @@
...
+			if vaultAuthRole != "" {
+				c.Command = append(c.Command, fmt.Sprintf("--auth-backend-role=%s", vaultAuthRole))
+			}
+			if configFile != "" {
+				podLogger.Log("msg", fmt.Sprintf("mutating %s with config file: %s", container.Name, configFile))
+				c.Command = append(c.Command, fmt.Sprintf("--config-file=%s", configFile))
+			} else {
+				podLogger.Log("msg", fmt.Sprintf("mutating %s without config file", container.Name))
+			}
+			c.Command = append(c.Command, fmt.Sprintf("--command=%s", originalCommand))
```
benwh: TODO: Test whether this correctly preserves spaces in commands

pkg/vault/envconsul/webhook.go
```diff
@@ -0,0 +1,177 @@
...
+	if !ok {
+		podLogger.Log("msg", "skipping: no configuration annotation")
+		return admission.PatchResponse(pod, pod)
+	}
+
+	vaultConfig := &corev1.ConfigMap{}
+	i.client.Get(ctx, i.vaultConfigKey, vaultConfig)
+	vaultAddress, _ := vaultConfig.Data["vault_address"]
+	vaultAuthMountPath, _ := vaultConfig.Data["auth_mount_path"]
+	vaultAuthRole, _ := vaultConfig.Data["auth_role"] // TODO: Add to the config file
```
benwh: We can drop this TODO right?

pkg/vault/envconsul/webhook.go
benwh: Are we double-appending here?

Would recommend only reassigning `c.Command` in one place.

pkg/vault/envconsul/webhook.go
```diff
@@ -0,0 +1,177 @@
...
+			VolumeMounts: getVolumeMounts(path),
+		},
+	}
+}
+
+func getVolumeMounts(path string) []corev1.VolumeMount {
+	return []corev1.VolumeMount{
+		{
+			Name:      "vault",
+			MountPath: path,
```
benwh: Could add `ReadOnly: true` here

pkg/vault/envconsul/webhook.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+}
+
+func getVolumes() []corev1.Volume {
+	return []corev1.Volume{
+		{
+			Name: "vault",
+			VolumeSource: corev1.VolumeSource{
+				EmptyDir: &corev1.EmptyDirVolumeSource{
+					Medium: corev1.StorageMediumMemory,
```
benwh: Do we actually want these stored in memory? I would have thought it'd be better to default to the node's default, which will be SSD in our case.
No use in wasting RAM to store binaries that'll only be called once?

pkg/vault/envconsul/webhook.go
```diff
@@ -0,0 +1,177 @@
...
+	config, ok := pod.Annotations[fmt.Sprintf("%s/configs", name)]
+	if !ok {
+		podLogger.Log("msg", "skipping: no configuration annotation")
+		return admission.PatchResponse(pod, pod)
+	}
+
+	vaultConfig := &corev1.ConfigMap{}
+	i.client.Get(ctx, i.vaultConfigKey, vaultConfig)
+	vaultAddress, _ := vaultConfig.Data["vault_address"]
+	vaultAuthMountPath, _ := vaultConfig.Data["auth_mount_path"]
```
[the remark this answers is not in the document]
dyson: Yeah we should check why Lawrence added this.

pkg/vault/envconsul/webhook.go
benwh: Actually this looks fine, will test this later though

benwh: Actually this is fine, ignore!

pkg/vault/envconsul/webhook.go
dyson: Good catch, this could be wasteful given enough pods on a node!

cmd/vault-manager/main.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/gocardless/theatre/pkg/signals"
+	"github.com/gocardless/theatre/pkg/vault/envconsul"
+)
+
+var (
+	app                       = kingpin.New("vault-manager", "Manages vault.crd.gocardless.com resources").Version(Version)
+	namespace                 = app.Flag("namespace", "Kubernetes webhook service namespace").Default("theatre-system").String()
+	serviceName               = app.Flag("service-name", "Webhook service name").Default("theatre-vault-manager").String()
+	webhookName               = app.Flag("webhook-name", "Mutating webhook name").Default("theatre-vault").String()
+	webhookConfigMapName      = app.Flag("webhook-configmap-name", "Webhook configMap name containing theatre-envconsul configuration").Default("theatre-vault-manager-webhook").String()
```
lawrencejones: I commented in the other PR, I think we can and should remove this.

We shouldn't need any configuration, and it makes things much simpler if we can just take flags set by environment variables, for example.

https://github.com/gocardless/theatre/pull/79/files#r350658999

pkg/vault/envconsul/webhook.go
```diff
@@ -0,0 +1,196 @@
...
+	return corev1.VolumeMount{
+		Name:      "vault",
+		MountPath: path,
+		ReadOnly:  readOnly,
+	}
+}
+
+func buildVolume() corev1.Volume {
+	return corev1.Volume{
+		Name: "vault",
```
lawrencejones: We should be super careful with these names and make sure they are very explicit. Otherwise we run the risk of colliding with people legitimately creating volume mounts.

This may well collide with our vault statefulsets, for example. And that collision would be confusing and disruptive.

cmd/vault-manager/main.go
jace-ys: Have removed this now and accept flags for `--theatre-image` and `--install-path`

PR #75, opened 2019-11-21
Opt-out from metric flags

lawrencejones:
For single-use commands, we probably don't want metrics. But we do want
the common logging configuration.

Allow people to opt-out of the metrics flags when appropriate.

### DIFF
--- a/cmd/rbac-manager/main.go
+++ b/cmd/rbac-manager/main.go
@@ -27,7 +27,7 @@ var (
 	app     = kingpin.New("rbac-manager", "Manages rbac.crd.gocardless.com resources").Version(Version)
 	refresh = app.Flag("refresh", "Refresh interval checking directory sources").Default("1m").Duration()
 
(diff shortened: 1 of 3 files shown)


cmd/rbac-manager/main.go
```diff
@@ -27,7 +27,7 @@ var (
 	app     = kingpin.New("rbac-manager", "Manages rbac.crd.gocardless.com resources").Version(Version)
 	refresh = app.Flag("refresh", "Refresh interval checking directory sources").Default("1m").Duration()
 
-	commonOpts = cmd.NewCommonOptions(app)
+	commonOpts = cmd.NewCommonOptions(app, true)
```
dyson: Is it an anti-pattern to do this? 

```
exposeMetrics := true
commonOpts = cmd.NewCommonOptions(app, exposeMetrics)
```

lawrencejones: You can make this more explicit if you want. I'll help you refactor in a bit

PR #79, opened 2019-11-25
Config for vault-manager

jace-ys:
This adds the deployment config for the vault-manager and related
resources. The vault-manager is deployed as a StatefulSet as we only
want a single replica of it.

We use Kustomize vars to grab the deployment image and pass it to
vault-manager, which runs the mutating webhook that injects the init
container running this same image.

### DIFF
--- a/config/base/managers/vault.yaml
+++ b/config/base/managers/vault.yaml
@@ -1,0 +1,110 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: vault-manager
+  labels:
+    group: vault.crd.gocardless.com
+rules:
+  - apiGroups:
+      - ""
+      - admissionregistration.k8s.io
+    resources:
+      - services
+      - mutatingwebhookconfigurations
+    verbs:
+      - "*"
+  - apiGroups:
+      - ""
+    resources:
+      - configmaps
+    verbs:
+      - list
+      - watch
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: vault-manager
+  labels:
+    group: vault.crd.gocardless.com
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: vault-manager
+  labels:
+    group: vault.crd.gocardless.com
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: vault-manager
+subjects:
+  - kind: ServiceAccount
+    name: vault-manager
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: vault-manager-webhook
+  labels:
+    group: vault.crd.gocardless.com
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - configmaps
+    resourceNames:
+      - theatre-vault-manager-webhook
+    verbs:
+      - get
+---
+apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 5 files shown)


config/base/managers/vault.yaml
```diff
@@ -0,0 +1,119 @@
...
+  template:
+    metadata:
+      labels:
+        group: vault.crd.gocardless.com
+    spec:
+      serviceAccountName: vault-manager
+      terminationGracePeriodSeconds: 10
+      containers:
+        - command:
+            - /vault-manager
```
lawrencejones: This should be /usr/local/bin/vault-manager

config/base/managers/vault.yaml
```diff
@@ -0,0 +1,119 @@
...
+          image: eu.gcr.io/gc-containers/gocardless/theatre:latest
+          imagePullPolicy: Always
```
lawrencejones: Always is an odd default- let's remove this, otherwise every time the pod boots we'll pull an image.

Fwiw, if the tag was `latest`, by default we would have `Always` as a pull policy. But we'll be overriding this tag with kustomize, and you normally want the default pull policy of `IfNotPresent`

config/base/managers/vault.yaml
```diff
@@ -0,0 +1,119 @@
...
+          env:
+            - name: POD_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+          resources:
+            limits:
+              cpu: 500m
+              memory: 256Mi
+---
```
lawrencejones: I don't think we need this config. The value for the theatre image can be passed into our manager using the downward API ([link](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#the-downward-api)), something like this:

```yaml
env:
  - name: THEATRE_IMAGE
    valueFrom:
      resourceFieldRef:
        containerName: manager
        resource: image
```

This happens to be much more sensible too, as if you ever change the interface of theatre-envconsul, you won't have a mutating webhook stamping out the old interface with a new image or vise versa. It's much saner for us to ensure the webhook provisions theatre-envconsul's of the same exact version.

config/base/managers/vault.yaml
```diff
@@ -0,0 +1,119 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: vault-manager-webhook
+  labels:
+    group: vault.crd.gocardless.com
+data:
+  image: eu.gcr.io/gc-containers/gocardless/theatre:7af2a24e5739912fb5fadb0749b96fab5ef88846
+  install_path: /run/vault
```
lawrencejones: Then the install path we can leave as the default, I think. No reason for us to have any configmap at all.

config/base/managers/vault.yaml
jace-ys: changed it thanks!

config/base/managers/vault.yaml
jace-ys: have removed the configMap and added this in

config/base/managers/vault.yaml
jace-ys: have removed it here, but just to highlight that `imagePullPolicy: Always` is used in the other managers, rbac/workloads so that might need changing too

config/base/managers/vault.yaml
jace-ys: @lawrencejones seems like this wouldn't work. the `image` field isn't exposed via the downward API. looking at an alternative for this 😢

jace-ys: Kustomize vars works here, so we're using that instead

PR #81, opened 2019-11-25
Acceptance tests for theatre-envconsul

lawrencejones:
Provision a Vault instance inside the Kind cluster. We can then use this
instance to validate the behaviour of theatre-envconsul by running a pod
with a specific service account that outputs its environment.

We don't yet test the configuration file for theatre-envconsul, nor the
mutating webhook. It would make sense to add these later.

To get this working, I needed to amend our acceptance environment to
boot a cluster using a custom kind-e2e.yaml config. While there, I've
changed the dockerfile we use to generate a test image to use what we
will for production, as this docker image contains things other than
just our binaries now (envconsul). This comes at the cost of being
slower to build, but is the correct way to do things.

Configuring Vault with appropriate policies is done using
theatre-envconsul itself, as the acceptance tests aren't guaranteed to
be run within the cluster where the service is accessible. It might make
sense for us to vendor a terraform file that can provide this
configuration in future, something that matches what we use internally
at GC.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 FROM golang:1.13.4 as builder
 WORKDIR /go/src/github.com/gocardless/theatre
 COPY . /go/src/github.com/gocardless/theatre
-RUN CGO_ENABLED=0 make VERSION=$(cat VERSION) build
+RUN make VERSION=$(cat VERSION) build
 
 # Use ubuntu as our base package to enable generic system tools
 FROM ubuntu:bionic-20190807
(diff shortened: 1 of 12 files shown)


Dockerfile
```diff
@@ -2,7 +2,7 @@
 FROM golang:1.13.4 as builder
 WORKDIR /go/src/github.com/gocardless/theatre
 COPY . /go/src/github.com/gocardless/theatre
-RUN CGO_ENABLED=0 make VERSION=$(cat VERSION) build
+RUN make VERSION=$(cat VERSION) build
```
lawrencejones: CGO_ENABLED is already set in the makefile

Dockerfile
```diff
@@ -26,9 +26,9 @@ RUN set -x \
     && curl -o /tmp/envconsul.zip -fsL https://releases.hashicorp.com/envconsul/0.9.1/envconsul_${ENVCONSUL_VERSION}_linux_amd64.zip \
     && echo ${ENVCONSUL_SHA256} /tmp/envconsul.zip | sha256sum -c \
     && unzip /tmp/envconsul.zip -d /tmp \
-    && mv /tmp/envconsul / \
+    && mv /tmp/envconsul /usr/local/bin/ \
```
lawrencejones: Let's put everything in /usr/local/bin so it's in the normal path.

PR #82, opened 2019-11-25
Disable vault retries in envconsul

jace-ys:
This should disable vault retries in envconsul by setting `enabled: false` in the generated config file

### DIFF
--- a/cmd/theatre-envconsul/main.go
+++ b/cmd/theatre-envconsul/main.go
@@ -455,6 +455,9 @@ func (o *vaultOptions) EnvconsulConfig(env environment, token, command string) *
 		Vault: envconsulVault{
 			Address: o.Address,
 			Token:   token,
+			Retry: envconsulRetry{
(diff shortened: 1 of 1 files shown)


lawrencejones:
Hey @jace-ys, I've rebased on master to pick-up the acceptance tests. If this goes green, we know envconsul has worked end-to-end and we should be safe to merge things to master.

I've also added a commit message: as a rule of thumb, we try to be detailed about why we're doing things in our commit history. Hopefully the new message can be a good example of what we normally look for!

jace-ys:
> Hey @jace-ys, I've rebased on master to pick-up the acceptance tests. If this goes green, we know envconsul has worked end-to-end and we should be safe to merge things to master.
> 
> I've also added a commit message: as a rule of thumb, we try to be detailed about why we're doing things in our commit history. Hopefully the new message can be a good example of what we normally look for!

Thanks, will keep that in mind !

PR #83, opened 2019-11-26
Accept command as args in theatre-envconsul

jace-ys:
The command to be executed by `theatre-envconsul` is currently taken in as a flag `--command`. This makes the interface neater by accepting the command as arguments instead, as that's the command we want theatre-envconsul to exec. Have also updated the mutating webhook to reflect this change.

### DIFF
--- a/cmd/theatre-envconsul/main.go
+++ b/cmd/theatre-envconsul/main.go
@@ -43,8 +43,8 @@ var (
 	exec             = app.Command("exec", "Authenticate with vault and exec envconsul")
 	execVaultOptions = newVaultOptions(exec)
 	execConfigFile   = exec.Flag("config-file", "app config file").String()
-	execCommand      = exec.Flag("command", "Command to execute").Required().String()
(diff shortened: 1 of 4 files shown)


jace-ys:
Waiting for #73 to be merged, and I'll rebase this while adding changes to the webhook in this PR

lawrencejones:
@jace-ys you should be clear to rebase this now

cmd/theatre-envconsul/main.go
```diff
@@ -43,8 +43,8 @@ var (
 	exec             = app.Command("exec", "Authenticate with vault and exec envconsul")
 	execVaultOptions = newVaultOptions(exec)
 	execConfigFile   = exec.Flag("config-file", "app config file").String()
-	execCommand      = exec.Flag("command", "Command to execute").Required().String()
 	execInstallPath  = exec.Flag("install-path", "Path containing installed binaries").Default(defaultInstallPath).String()
+	execCommand      = exec.Arg("command", "Command to execute").Required().String()
```
lawrencejones: I think you need to make this `Strings()` right? So that it's an array?

If you do that, compiler errors should help you figure out what remains to be changed

PR #86, opened 2019-11-28
Increase vault-manager replicas

jace-ys:
We need to run more replicas of the vault-manager to ensure high availability,
as we need the mutating webhook to be always available - new pods will not be
created otherwise. If the manager ever needs to manage custom resources in the
future, we need to implement leader election in vault-manager to prevent races.

### DIFF
--- a/config/base/managers/vault.yaml
+++ b/config/base/managers/vault.yaml
@@ -81,6 +81,10 @@ metadata:
     group: vault.crd.gocardless.com
 spec:
   serviceName: vault-manager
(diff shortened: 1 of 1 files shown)


dyson:
This seems to work fine. Registering the webhook multiple times seems to be ok - only the one service is created as expected, and the certificate for the webhooks are self signed so there is no conflict there. If me move over to using cert-manager to handle all off our certs, we might have an issue with this, but for now it looks like it is ok. We will keep testing this however so if there is any issue we can revisit this.

PR #93, opened 2019-12-03
Ephemeral service account tokens for Vault

lawrencejones:
Kubernetes service account tokens, by default, are not rotated. Our
Vault authentication chain requires that Kubernetes pods send their
tokens to the Vault server, outside of the Kubernetes cluster.

If a token was to be compromised in that exchange, the attacker would be
able to impersonate this Kubernetes service account indefinitely. This
is a significant security risk.

This commit changes our mutating webhook to mount service account
credentials using the kubelet projected volume mounts API. This API uses
the kubelet to inject ephemeral service account tokens that are rotated
periodically into the containers. We can use these tokens to
authenticate against Vault, instead of tokens that are never revoked.

In addition, we can set an audience for the projected token to be
something other than the Kubernetes API server. This would prevent any
leaked tokens being used for anything other than authenticating with
Vault. While we don't use this at present, as our Vault installation
relies on using this token to reflect token review requests back to the
API server, we may switch to this in future if we think the security
benefit is compelling.


cmd/theatre-envconsul/acceptance/acceptance.go
```diff
@@ -190,9 +173,75 @@ func (r *Runner) Prepare(logger kitlog.Logger, config *rest.Config) error {
...
+        - --install-path=/usr/local/bin
+        - --service-account-token-file=/var/run/secrets/kubernetes.io/vault/token
+        - --
+        - env
+      volumeMounts:
+        - name: theatre-envconsul-serviceaccount
+          mountPath: /var/run/secrets/kubernetes.io/vault
+`
+
+const annotatedPodYAML = `
```
lawrencejones: CC @dyson @jace-ys

cmd/theatre-envconsul/acceptance/acceptance.go
```diff
@@ -190,9 +173,75 @@ func (r *Runner) Prepare(logger kitlog.Logger, config *rest.Config) error {
...
+metadata:
+  generateName: read-a-secret-
+  namespace: staging # provisioned by the acceptance kustomize overlay
+  annotations:
+    envconsul-injector.vault.crd.gocardless.com/configs: app
+spec:
+  serviceAccountName: secret-reader
+  restartPolicy: Never
+  containers:
+    - name: app
```
jace-ys: think this should be `print-env` to match the YAML above, but this shouldn't affect the test

lawrencejones: Yep good point, I'll change them to match (though I'll go with app in both cases, as that's what we like to use)

PR #94, opened 2019-12-03
Acceptance test runners, with optional prepare steps

lawrencejones:
For acceptance tests, it is good to be able to prepare some additional
resources prior to running the tests themselves. To do this, refactor
our existing acceptance test suites to use a Runner interface, where
each suite can provide optional prepare commands.

This allows an operator to run cmd/acceptance prepare to re-converge
their cluster, improving the local developer experience. Using this
pattern enables a refactor of lots of test-only functionality from the
theatre-envconsul binary, which is a win for readability.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,7 +46,7 @@ jobs:
       - checkout
       - run:
           name: Build test binaries
-          command: make build-linux
+          command: make build-linux bin/acceptance.linux_amd64
       - persist_to_workspace:
           root: /go/src/github.com/gocardless/theatre
           paths: ['bin']
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,3 +4,8 @@ tmp/
 .kube/
 bin/
 config/
+
+# Acceptance test suites don't need recompiling into our docker image
+cmd/acceptance/
+cmd/theatre-envconsul/acceptance/
+pkg/workloads/console/acceptance/
--- a/cmd/acceptance/main.go
+++ b/cmd/acceptance/main.go
@@ -7,10 +7,13 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"reflect"
 	"strings"
 	"testing"
 	"time"
 
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog"
 
 	"github.com/alecthomas/kingpin"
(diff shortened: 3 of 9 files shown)


.circleci/config.yml
```diff
@@ -46,7 +46,7 @@ jobs:
       - checkout
       - run:
           name: Build test binaries
-          command: make build-linux
+          command: make build-linux bin/acceptance.linux_amd64
```
lawrencejones: I removed the acceptance binary from the default linux targets, as you often don't want to compile this into the dockerfile and it adds another binary for compilation in your feedback cycle

.dockerignore
```diff
@@ -4,3 +4,8 @@ tmp/
 .kube/
 bin/
 config/
+
+# Acceptance test suites don't need recompiling into our docker image
+cmd/acceptance/
```
lawrencejones: Ignoring these means you can modify the acceptance test, run `go run cmd/acceptance/main.go prepare`, and your docker image is already cached.

cmd/acceptance/main.go
```diff
@@ -142,21 +173,19 @@ func main() {
 	}
 }
 
-var _ = Describe("Acceptance", func() {
```
lawrencejones: This diff is horrible, better to just view the file. I've inlined the describe block into the run command, no biggie.

cmd/theatre-envconsul/acceptance/acceptance.go
```diff
@@ -29,19 +35,6 @@ metadata:
 spec:
   serviceAccountName: secret-reader
   restartPolicy: Never
-  initContainers:
```
lawrencejones: We no longer require this init container as the work to provision vault is done within the prepare block

cmd/theatre-envconsul/acceptance/acceptance.go
```diff
@@ -60,13 +53,150 @@ spec:
...
+// - Mounts a kv2 secrets engine at secret/
+// - Creates a Kubernetes auth backend mounted at auth/kubernetes
+// - Configures the Kubernetes backend to authenticate against the currently detected
+//   Kubernetes API server (the current cluster, if run from within)
+// - For all successful Kubernetes logins, the user is assigned a token that maps to a
+//   cluster-reader policy, which permits reading of secrets from:
+//
+//   secret/data/kubernetes/{namespace}/{service-account-name}/*
+//
+func (r *Runner) Prepare(logger kitlog.Logger, config *rest.Config) error {
```
lawrencejones: This is lifted from the theatre-envconsul configure command

cmd/theatre-envconsul/main.go
```diff
@@ -162,127 +159,6 @@ func mainError(ctx context.Context, command string) (err error) {
 			return errors.Wrap(err, "failed to execute envconsul")
 		}
 
-	// This command should only be used for preparing a test environment. It is
```
lawrencejones: This is now the prepare in the acceptance test

config/overlays/acceptance/vault.yaml
```diff
@@ -24,9 +24,11 @@ metadata:
   name: vault
   namespace: vault
 spec:
+  type: NodePort
```
lawrencejones: Expose the vault service onto your local machine. You can now go to localhost:8200 when kind is running to access vault.

pkg/workloads/console/acceptance/acceptance.go
```diff
@@ -48,14 +45,20 @@ func newClient(kubeConfigPath string) clientset {
 	return clientset{Clientset: core, workloadsV1alpha1: workloadsClient}
 }
 
-func Run(logger kitlog.Logger, kubeConfigPath string) {
+type Runner struct{}
```
lawrencejones: Adjust the console acceptance tests to use the new runner format

cmd/acceptance/main.go
```diff
@@ -142,21 +173,19 @@ func main() {
...
-	ctx, cancel := signals.SetupSignalHandler()
-	defer cancel()
-
-	cfgPathBytes, err := exec.CommandContext(ctx, "kind", "get", "kubeconfig-path", "--name", contextName).Output()
-	cfgPath := string(bytes.TrimSpace(cfgPathBytes))
+func mustClusterConfig() *rest.Config {
+	config, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		clientcmd.NewDefaultClientConfigLoadingRules(),
+		&clientcmd.ConfigOverrides{
+			CurrentContext: "kind-e2e",
```
rnaveiras: This should be parametrized with the name of the cluster

cmd/acceptance/main.go
```diff
@@ -114,6 +129,13 @@ func main() {
 			app.Fatalf("failed to install manager: %v", err)
 		}
 
+		for _, runner := range Runners {
+			logger.Log("msg", "running prepare", "runner", reflect.TypeOf(runner).Elem().Name())
+			if err := runner.Prepare(logger, mustClusterConfig()); err != nil {
```
jace-ys: you could pull `mustClusterConfig()` outside the for loop, similar to how you did in `run`

cmd/theatre-envconsul/acceptance/acceptance.go
```diff
@@ -60,13 +53,150 @@ spec:
...
+
+	logger.Log("msg", "enabling auth mount", "path", AuthBackendMountPath, "options", enableOptions)
+	client.Sys().DisableAuth(AuthBackendMountPath)
+	if err := client.Sys().EnableAuthWithOptions(AuthBackendMountPath, enableOptions); err != nil {
+		return err
+	}
+
+	var ca string
+
+	if string(config.CAData) == "" {
```
jace-ys: would `if len(config.CAData) == 0` be better? it skips an additional type cast

cmd/theatre-envconsul/acceptance/acceptance.go
```diff
@@ -60,13 +53,150 @@ spec:
...
+	}
+
+	// We'll be running the acceptance tests from outside the kubernetes cluster, where the
+	// API server will have an IP address that is relative to the host machine. When we're
+	// within the cluster, like Vault, we want to talk to kubernetes.default.svc to ensure
+	// we're tapping the host IP address.
+	backendConfigPath := fmt.Sprintf("auth/%s/config", AuthBackendMountPath)
+	backendConfig := map[string]interface{}{
+		"kubernetes_host":    "https://kubernetes.default.svc",
+		"kubernetes_ca_cert": string(ca),
```
jace-ys: this seems like you're casting a string to a string. i would do `var ca []byte` and deal with bytes in the above section, only finally casting it to string here

lawrencejones: Yep true, the code is casting a few times here. I tend to try and avoid using byte slices where possible and to lean on strings, but that gets a bit messy in this instance.

Going to clean it up.

PR #98, opened 2019-12-05
acceptance --target

lawrencejones:
Optionally filter tests for when you're running acceptance tests
locally. Eventually we'll switch this to use plain old ginkgo, but for
now this makes life better.


lawrencejones:
`acceptance run run --target envconsul` for example

PR #100, opened 2019-12-05
pkg/vault/README.md

lawrencejones:
Aim to explain how our webhook works, what it does and the interaction
with Vault.


lawrencejones:
Merging as docs-only

PR #104, opened 2019-12-12
Use vault path prefix to support multi-cluster secrets

jace-ys:
To support using secrets across multiple clusters, the vault-prefixed secret key (ie.
`vault:<secret-path>/<secret-name>/2019121201`) should not specify the project being
targeted. Instead, this information needs to be populated at deploy time, using per cluster config.

To do this, the mutating webhook reads a secret mount prefix path from the specific cluster's vault-config ConfigMap, and injects the namespace and service account information, before passing that on to `theatre-envconsul` via a `--vault-path-prefix flag`. That path is prefixed to each secret name specified in the app config before writing to the envconsul config file.

This means that instead of the full path to the secret, the app config should now only supply the required secret name/version, ie. `vault:<secret-name>/2019121201`, which will be transformed into `<secret-path>/<secret-name>/2019121201` by the mutating webhook.

This ties in with https://github.com/gocardless/anu/pull/3329, where we provision the ConfigMap with an additional `secret_mount_path_prefix` and modify the interface of `anu secrets` to inform users of the new vault-prefixed secret key they should use.

### DIFF
--- a/cmd/theatre-envconsul/acceptance/acceptance.go
+++ b/cmd/theatre-envconsul/acceptance/acceptance.go
@@ -200,12 +200,13 @@ spec:
       imagePullPolicy: Never
       env:
         - name: VAULT_RESOLVED_KEY
-          value: vault:secret/data/kubernetes/staging/secret-reader/jimmy
+          value: vault:jimmy
       command:
         - /usr/local/bin/theatre-envconsul
       args:
         - exec
         - --vault-address=http://vault.vault.svc.cluster.local:8200
+        - --vault-path-prefix=secret/data/kubernetes/staging/secret-reader
         - --no-vault-use-tls
         - --install-path=/usr/local/bin
         - --service-account-token-file=/var/run/secrets/kubernetes.io/vault/token
@@ -234,7 +235,7 @@ spec:
       imagePullPolicy: Never
       env:
         - name: VAULT_RESOLVED_KEY
-          value: vault:secret/data/kubernetes/staging/secret-reader/jimmy
+          value: vault:jimmy
       command:
         - env
 `
--- a/config/overlays/acceptance/vault.yaml
+++ b/config/overlays/acceptance/vault.yaml
@@ -79,3 +79,4 @@ data:
   address: http://vault.vault.svc.cluster.local:8200
   auth_mount_path: kubernetes
   auth_role: default
+  secret_mount_path_prefix: secret/data/kubernetes
--- a/pkg/vault/envconsul/webhook_test.go
+++ b/pkg/vault/envconsul/webhook_test.go
@@ -35,9 +35,10 @@ var _ = Describe("PodInjector", func() {
 	BeforeEach(func() {
 		injector = &PodInjector{
 			VaultConfig: VaultConfig{
-				Address:       "https://vault.example.com",
-				AuthMountPath: "kubernetes.gc-prd-effc.cluster",
-				AuthRole:      "default",
+				Address:               "https://vault.example.com",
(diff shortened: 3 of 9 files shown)


cmd/theatre-envconsul/acceptance/acceptance.go
```diff
@@ -200,12 +200,13 @@ spec:
...
       env:
         - name: VAULT_RESOLVED_KEY
-          value: vault:secret/data/kubernetes/staging/secret-reader/jimmy
+          value: vault:jimmy
       command:
         - /usr/local/bin/theatre-envconsul
       args:
         - exec
         - --vault-address=http://vault.vault.svc.cluster.local:8200
+        - --vault-path-prefix=secret/data/kubernetes/staging/secret-reader
```
dyson: We should add the project to the prefix everywhere. It isn't strictly needed, but it better shows the intention of how you should use this.

config/overlays/acceptance/vault.yaml
```diff
@@ -79,3 +79,4 @@ data:
   address: http://vault.vault.svc.cluster.local:8200
   auth_mount_path: kubernetes
   auth_role: default
+  secret_mount_path_prefix: secret/data/kubernetes
```
dyson: Add project here as well.

pkg/vault/envconsul/webhook_test.go
```diff
@@ -35,9 +35,10 @@ var _ = Describe("PodInjector", func() {
 	BeforeEach(func() {
 		injector = &PodInjector{
 			VaultConfig: VaultConfig{
-				Address:       "https://vault.example.com",
-				AuthMountPath: "kubernetes.gc-prd-effc.cluster",
-				AuthRole:      "default",
+				Address:               "https://vault.example.com",
+				AuthMountPath:         "kubernetes.gc-prd-effc.cluster",
+				AuthRole:              "default",
+				SecretMountPathPrefix: "secret/data/kubernetes",
```
dyson: And here

pkg/vault/envconsul/webhook_test.go
```diff
@@ -121,6 +122,8 @@ var _ = Describe("PodInjector", func() {
 								"/var/run/theatre-envconsul",
 								"--vault-address",
 								"https://vault.example.com",
+								"--vault-path-prefix",
+								"secret/data/kubernetes/staging/secret-reader",
```
dyson: and project here

pkg/vault/envconsul/webhook_test.go
```diff
@@ -232,6 +235,8 @@ var _ = Describe("PodInjector", func() {
...
```
dyson: and finally project here!

dyson:
Looking good! I think instead of project we should call it `group` or `cluster_group`? I think being in Google Cloud project is already strongly associated with a Google Cloud project. And thinking about the terraform, using a project variable in the module is already taken, so would be confusing. 

Thoughts?

cmd/theatre-envconsul/acceptance/acceptance.go
jace-ys: Discussed this in person - theatre doesn't need to know about our vault setup, and can work with any path prefix.

jace-ys:
> Looking good! I think instead of project we should call it `group` or `cluster_group`? I think being in Google Cloud project is already strongly associated with a Google Cloud project. And thinking about the terraform, using a project variable in the module is already taken, so would be confusing.
> 
> Thoughts?

Discussed this in person - theatre doesn't need to know about our vault setup, and can work with any path prefix.

PR #111, opened 2019-12-18
Apply a default fsGroup in envconsul webhook

lawrencejones:
This permits us to read from the projected service account token volume
mount, which is otherwise restricted to root only, and therefore
unsuitable for containers running as non-root (which we'd prefer to
encourage).

### DIFF
--- a/pkg/vault/envconsul/webhook.go
+++ b/pkg/vault/envconsul/webhook.go
@@ -236,6 +236,17 @@ func (i PodInjector) Inject(pod corev1.Pod) *corev1.Pod {
 		},
 	)
 
+	// If we don't already have an fsGroup set, we'll need to configure it so we can read
+	// the contents of the volumes we mount. Failing to do this will prevent us from reading
+	// the projected service account token.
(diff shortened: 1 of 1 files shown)


pkg/vault/envconsul/webhook.go
```diff
@@ -236,6 +236,17 @@ func (i PodInjector) Inject(pod corev1.Pod) *corev1.Pod {
...
 	)
 
+	// If we don't already have an fsGroup set, we'll need to configure it so we can read
+	// the contents of the volumes we mount. Failing to do this will prevent us from reading
+	// the projected service account token.
+	if mutatedPod.Spec.SecurityContext == nil {
+		mutatedPod.Spec.SecurityContext = &corev1.PodSecurityContext{}
+	}
+	if mutatedPod.Spec.SecurityContext.FSGroup == nil {
+		var defaultFSGroup int64 = 1000
```
jace-ys: `defaultFSGroup := int64(1000)` seems more Go-ish

PR #118, opened 2020-01-02
envconsul fork for duplicate secret env

lawrencejones:
https://github.com/hashicorp/envconsul/pull/231

Incorporate the changes from the linked PR that enable support for multiple
secret values with the same vault prefix.


lawrencejones:
This appears to work, having now deployed it. +1 @dyson?

PR #132, opened 2020-04-03
Create CLI for creating consoles

jackatbancast:
This change introduces a `theatre-consoles` kingpin application.
The application has a single command, ``create`` which will create a
console instance, ready to attach to.

This is part of a series of changes to provide a command line to
work with consoles in theatre. The management logic currently in the
consoles application will likely move to be closer to the runner in
the near future.

### DIFF
--- a/cmd/theatre-consoles/create.go
+++ b/cmd/theatre-consoles/create.go
@@ -1,0 +1,47 @@
+package main
+
+import (
+	"context"
+
+	kitlog "github.com/go-kit/kit/log"
+	"github.com/gocardless/theatre/pkg/workloads/console/runner"
+	consoleRunner "github.com/gocardless/theatre/pkg/workloads/console/runner"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+	createSelector = create.Flag("selector", "Selector that matches console template").Required().String()
+	createTimeout  = create.Flag("timeout", "Timeout for the new console").Duration()
+	createReason   = create.Flag("reason", "Reason for creating console").String()
+	createCommand  = create.Arg("command", "Command to run in console").Strings()
+)
+
+func Create(ctx context.Context, logger kitlog.Logger, runner *runner.Runner) error {
+	var err error
+
+	// Create and attach to the console
+	tpl, err := runner.FindTemplateBySelector(metav1.NamespaceAll, *createSelector)
+	if err != nil {
+		return err
+	}
+
+	opt := consoleRunner.Options{Cmd: *createCommand, Timeout: int(createTimeout.Seconds()), Reason: *createReason}
+	csl, err := runner.Create(tpl.Namespace, *tpl, opt)
+	if err != nil {
+		return nil
+	}
+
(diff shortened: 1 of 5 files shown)


cmd/theatre-consoles/create.go
```diff
@@ -0,0 +1,47 @@
...
+	createTimeout  = create.Flag("timeout", "Timeout for the new console").Duration()
+	createReason   = create.Flag("reason", "Reason for creating console").String()
+	createCommand  = create.Arg("command", "Command to run in console").Strings()
+)
+
+func Create(ctx context.Context, logger kitlog.Logger, runner *runner.Runner) error {
+	var err error
+
+	// Create and attach to the console
+	tpl, err := runner.FindTemplateBySelector(metav1.NamespaceAll, *createSelector)
```
benwh, line 23: I think that this only works if the user has permission to query ConsoleTemplates at the cluster scope, which not everyone/everything (service accounts) will have. I suspect we'll need to add namespace as well to get around this, but happy to leave that until a later PR.

benwh: Actually yes, I think we definitely need to be able to supply namespace, in order to be able to address console templates which exist in multiple namespaces with the same label sets:

```console
$ go run ./... create --selector 'app=etcd'
consoles: error: unexpected error: expected to discover 1 console template, but actually found: [platform/etcd-cluster-default production/etcd-cluster-default staging/etcd-cluster-default]
```

cmd/theatre-consoles/main.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+}
+
+func Run(ctx context.Context, logger kitlog.Logger) error {
+	config, err := kconf.GetConfig()
+	if err != nil {
+		return err
+	}
+
+	client, err := kubernetes.NewForConfig(config)
```
benwh, line 48: Personally I'd like to have the kube context parameterisable here. Otherwise it becomes a bit awkward if you need to boot a console in two or more clusters, needing to switch your context around. WDYT?

cmd/theatre-consoles/create.go
```diff
@@ -0,0 +1,47 @@
...
+	if err != nil {
+		return nil
+	}
+
+	pod, err := runner.GetAttachablePod(csl)
+	if err != nil {
+		return nil
+	}
+
+	logger.Log("pod", pod.Name, "msg", "console pod created")
```
benwh, line 44: I assume that attaching to the console will be part of a subsequent PR?

dyson:
LGTM with the bits Ben addressed 👍

cmd/theatre-consoles/create.go
jackatbancast: Yep, the thinking is that it'd probably be useful to chain this onto the attach command once we've implemented that here.

cmd/theatre-consoles/main.go
jackatbancast: Reflecting on this, I think it would be useful to parametrise this.
It would also be useful in the case you wanted to script anything using this cli.

PR #133, opened 2020-04-06
Add authorisation support to console CRDs

dyson:
Adds fields to the ConsoleTemplate CRD that facilitates implementing
peer-reviewed consoles.

`authorisationRules` provides a list of rules that commands can be
matched against and if a match is found, the console requires
authorisation to run from a defined number of subjects. It is intended
that the rules will be matched from top to bottom and the first match is
used.

`defaultAuthorisationRule` provides a default rule to match on if no
other rules match.

Both these fields are optional allowing for backward compatibility.

An additional CRD has also been added to facilitate the authorisation
process. The `ConsoleAuthorisation` kind provides a way to monitor
approvals provided to a console that matches a rule, or the default
rule.

The actual implementation of the peer-reviewed console feature is to
follow.


dyson:
FYI the only file manually modified in this PR is `pkg/apis/workloads/v1alpha1/console_types.go`.

All other files were created by running:
```
$ make codegen
...
$ make manifests
```

pkg/apis/workloads/v1alpha1/console_types.go
```diff
@@ -120,6 +120,28 @@ type ConsoleTemplateSpec struct {
 	// +kubebuilder:validation:Minimum=0
 	// +kubebuilder:validation:Maximum=604800
 	DefaultTTLSecondsAfterFinished *int32 `json:"defaultTtlSecondsAfterFinished,omitempty"`
+
+	// List of authorisation rules to match against in order from top to bottom.
+	// +optional
+	AuthorisationRules []ConsoleAuthorisationRule `json:"autorisationRules,omitempty"`
```
jackatbancast, line 126:
```suggestion
	AuthorisationRules []ConsoleAuthorisationRule `json:"authorisationRules,omitempty"`
```

pkg/apis/workloads/v1alpha1/console_types.go
```diff
@@ -120,6 +120,37 @@ type ConsoleTemplateSpec struct {
 	// +kubebuilder:validation:Minimum=0
 	// +kubebuilder:validation:Maximum=604800
 	DefaultTTLSecondsAfterFinished *int32 `json:"defaultTtlSecondsAfterFinished,omitempty"`
+
+	// List of authorisation rules to match against in order from top to bottom.
+	// +optional
+	AuthorisationRules []ConsoleAuthorisationRule `json:"authorisationRules,omitempty"`
+	// Default authorisation rule to use if no authorisationRules are defined or no authorisationRules match.
```
benwh, line 127:
```suggestion
	// Default authorisation rule to use if no authorisation rules are defined or no authorisation rules match.
```
Maybe nitpicky but I think it reads better if we don't refer to the exact field names.

pkg/apis/workloads/v1alpha1/console_types.go
```diff
@@ -120,6 +120,37 @@ type ConsoleTemplateSpec struct {
...
+	// +optional
+	DefaultAuthorisationRule *ConsoleDefaultAuthorisationRule `json:"defaultAuthorisationRule,omitempty"`
+}
+
+// ConsoleAuthorisationRule declares rules specifying what commands need to be authorised and by whom.
+type ConsoleAuthorisationRule struct {
+	Name string `json:"name"`
```
benwh, line 134: Missing a comment on this one? It's probably important given its ambiguity otherwise

pkg/apis/workloads/v1alpha1/console_types.go
```diff
@@ -120,6 +120,37 @@ type ConsoleTemplateSpec struct {
...
+	// List of subjects that can provide authorisation for the matched console command to run.
+	Subjects []rbacv1.Subject `json:"subjects"`
+}
+
+// ConsoleDefaultAuthorisationRule declares the default authorisation required for any commands that don't match any authorisationRules or when there are no authorisationRules.
+type ConsoleDefaultAuthorisationRule struct {
+	// The number of authorisations required from members of the subjects before the console can run.
+	AuthorisationsRequired int `json:"authorisationsRequired"`
+
+	// List of subjects that can provide authorisation for the matched console command to run.
```
benwh, line 152:
```suggestion
	// List of subjects that can provide authorisation for the console command to run.
```

pkg/apis/workloads/v1alpha1/console_types.go
```diff
@@ -131,3 +162,38 @@ type ConsoleTemplateList struct {
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec ConsoleAuthorisationSpec `json:"spec"`
+}
+
+// ConsoleAuthorisationSpec defines the specification for a console authorisation
+type ConsoleAuthorisationSpec struct {
+	// The reference to the console by name that this ConsoleAuthorisation belongs to.
+	ConsoleRef struct {
```
benwh, line 180: Should we not be using `corev1.LocalObjectReference` here?

See example earlier in the file

pkg/apis/workloads/v1alpha1/console_types.go
```diff
@@ -131,3 +162,38 @@ type ConsoleTemplateList struct {
...
+		Name string `json:"name"`
+	} `json:"consoleRef"`
+
+	// The user who created the referenced console.
+	Owner string `json:"owner"`
```
benwh, line 185: I'm not sure if we do actually want this - I think we originally put it in the CRD sketch so as to save a lookup to the referenced object and make the code a bit cleaner - but I reckon let's keep it for now and see how it goes?

pkg/apis/workloads/v1alpha1/console_types.go
```diff
@@ -120,6 +120,37 @@ type ConsoleTemplateSpec struct {
...
+// ConsoleAuthorisationRule declares rules specifying what commands need to be authorised and by whom.
+type ConsoleAuthorisationRule struct {
+	Name string `json:"name"`
+
+	// The match expression to compare to the command and arguments of the console.
+	// Uses the glob format https://github.com/gobwas/glob.
+	MatchCommand string `json:"matchCommand"`
+
+	// The number of authorisations required from members of the subjects before the console can run.
+	AuthorisationsRequired int `json:"authorisationsRequired"`
```
benwh, line 141: What strikes me here is that we've duplicated these two fields between `ConsoleAuthorisationRule` and `ConsoleDefaultAuthorisationRule`.

Is it possible for these to be extracted out into a struct, which we embed within `ConsoleAuthorisationRule`? (And if we did that, we might want to change the naming a bit).

I can't remember if it's doable in this world with all of the codegen, or not.

If we can do it then I think it might make our lives simpler when coming to write the reconciliation logic, as we'll probably need to be appending the default rule onto the end of the list of rules in order to process them, so it may make sense to use the same types.

dyson: Going to extract this into a struct called `ConsoleAuthorisers` and see how it goes.

pkg/apis/workloads/v1alpha1/console_types.go
dyson: Yeah totally correct. We might not find this useful and remove this later on. Let's see how it plays out.

pkg/apis/workloads/v1alpha1/console_types.go
dyson: 👍

pkg/apis/workloads/v1alpha1/console_types.go
dyson: Ah yep, will add this is is shown in audit logs in a comment to direct people on its use.

pkg/apis/workloads/v1alpha1/console_types.go
dyson: FYI had to add a json tag (`` `json:",inline"` ``) to the embedded struct for the codegen to work as expected.

PR #139, opened 2020-04-09
Add reconciliation skeleton for console authorisation

benwh:
a989a19: Make webhook struct fields non-public

This clears up the package namespace, and prevents a conflict that will
occur in a future commit!

70db410: Don't require update to authorisers in webhook

We discovered, while writing reconciliation code, that if you were to
perform an update against this resource but not add any authorisers,
then the update would be rejected.

This is unintuitive, it should be possible to perform operations such
as adding annotations or altering non-core fields, without any issue.

Change the validation code to only throw an error if there's greater
than 1 addition, as opposed to 0 or >1 additions.

030904f: Create authorisation object for consoles

When a console is created, that is based off a template which includes
authorisation rules, stamp out a `consoleauthorisation` object that will
be used by approvers to authorise the console instance.

This commit adds the reconciliation around the authorisation objects,
and skeleton tests, but leaves the implementation of determining whether
a console is authorised for a subsequent PR.

### DIFF
--- a/pkg/vault/envconsul/webhook.go
+++ b/pkg/vault/envconsul/webhook.go
@@ -29,7 +29,7 @@ const EnvconsulInjectorFQDN = "envconsul-injector.vault.crd.gocardless.com"
 
 func NewWebhook(logger kitlog.Logger, mgr manager.Manager, injectorOpts InjectorOptions, opts ...func(*admission.Handler)) (*admission.Webhook, error) {
 	var handler admission.Handler
-	handler = &Injector{
+	handler = &injector{
 		logger:  kitlog.With(logger, "component", "EnvconsulInjector"),
 		decoder: mgr.GetAdmissionDecoder(),
 		client:  mgr.GetClient(),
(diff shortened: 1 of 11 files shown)


pkg/vault/envconsul/webhook.go
```diff
@@ -180,34 +180,34 @@ func (i *Injector) Handle(ctx context.Context, req types.Request) (resp types.Re
...
 
-func newVaultConfig(cfgmap *corev1.ConfigMap) (VaultConfig, error) {
-	var cfg VaultConfig
+func newVaultConfig(cfgmap *corev1.ConfigMap) (vaultConfig, error) {
+	var cfg vaultConfig
 	return cfg, mapstructure.Decode(cfgmap.Data, &cfg)
 }
 
-// PodInjector isolates the logic around injecting theatre-envconsul away from anything to
+// nodInjector isolates the logic around injecting theatre-envconsul away from anything to
```
dyson, line 200: *podInjector

dyson:
Looks like a solid base to me! We can change anything we need to as we implement all the other logic but I reckon we are on a good path with this.

FYI I just force pushed to fix a single typo.

PR #141, opened 2020-04-14
Consoles refactor subcommands

jackatbancast:
This change has 2 stages.

The first removes the dependance of the consoles subcommands from flags, allowing them to be more easily used directly from other codebases, without having to add an integration layer over a kingpin application.

The second stage moves these function to be methods of option objects. This allows for a nicer builder pattern wehn integrating into external codebases.


jackatbancast:
Closed by #139, #140

PR #142, opened 2020-04-15
Fix recreation of console jobs on delete

dyson:
commit 4e16a56bc33ee626fc7637ffd64810e9ba01ecfa

    Fix console job name truncate length

    Jobs names are restricted to 63 characters as they are added as a label
    to the pods they create (and labels have a 63 character length
    restriction).

    We incorrectly limited job names thinking that the pods they created had
    a name limit of 63 characters. This fix gets us a few characters back in
    the job name and ensures our labels are the correct length.

commit e0752e6d460aa670eb9b38625d3b90688a5b7893

    Only create jobs on console creation and mark a console as destroyed if
    a console job has been deleted from underneath the console.

    Before this fix, deleting a job (even if the console had completed and
    was waiting for GC) would result in a new job being created for the
    console and therefore a new console pod executing.

    The recreation of the job is unexpected behaviour and this change
    ensures this double running cannot happen.

### DIFF
--- a/pkg/workloads/console/acceptance/acceptance.go
+++ b/pkg/workloads/console/acceptance/acceptance.go
@@ -175,6 +175,55 @@ func (r *Runner) Run(logger kitlog.Logger, config *rest.Config) {
 			}).Should(HaveOccurred(), "expected not to find console, but did")
 		})
 
+		Specify("Deleting a job", func() {
+			By("Create a console template")
+			template := buildConsoleTemplate(nil, false)
+			template, err := client.WorkloadsV1Alpha1().ConsoleTemplates(namespace).Create(template)
+			Expect(err).NotTo(HaveOccurred(), "could not create console template")
+
+			defer func() {
+				By("(cleanup) Delete the console template")
+				policy := metav1.DeletePropagationForeground
+				err = client.WorkloadsV1Alpha1().ConsoleTemplates(namespace).
+					Delete(templateName, &metav1.DeleteOptions{PropagationPolicy: &policy})
+				Expect(err).NotTo(HaveOccurred(), "could not delete console template")
+
+				Eventually(func() error {
+					_, err = client.WorkloadsV1Alpha1().ConsoleTemplates(namespace).Get(templateName, metav1.GetOptions{})
+					return err
+				}).Should(HaveOccurred(), "expected console template to be deleted, it still exists")
+			}()
+
+			By("Create a console")
+			console := buildConsole()
+			console, err = client.WorkloadsV1Alpha1().Consoles(namespace).Create(console)
+			Expect(err).NotTo(HaveOccurred(), "could not create console")
+
+			By("Expect a job has been created")
+			Eventually(func() error {
+				_, err = client.BatchV1().Jobs(namespace).Get(jobName, metav1.GetOptions{})
+				return err
+			}).ShouldNot(HaveOccurred(), "could not find job")
+
+			By("Delete the job")
+			policy := metav1.DeletePropagationForeground
+			err = client.BatchV1().Jobs(namespace).Delete(jobName, &metav1.DeleteOptions{PropagationPolicy: &policy})
+			Expect(err).NotTo(HaveOccurred(), "could not delete console job")
+
+			By("Expect that the job no longer exists")
+			Eventually(func() error {
+				_, err = client.BatchV1().Jobs(namespace).Get(consoleName, metav1.GetOptions{})
+				return err
+			}).Should(HaveOccurred(), "expected not to find job, but did")
+
+			By("Expect the console phase is Destroyed")
+			Eventually(func() workloadsv1alpha1.ConsolePhase {
+				console, err = client.WorkloadsV1Alpha1().Consoles(namespace).Get(consoleName, metav1.GetOptions{})
+				Expect(err).NotTo(HaveOccurred(), "could not find console")
+				return console.Status.Phase
+			}).Should(Equal(workloadsv1alpha1.ConsoleDestroyed))
+		})
+
 		Specify("Authorised console", func() {
 			By("Create a console template")
 			var ttl int32 = 30
--- a/pkg/workloads/console/controller.go
+++ b/pkg/workloads/console/controller.go
@@ -501,10 +501,7 @@ func (r *reconciler) buildJob(template *workloadsv1alpha1.ConsoleTemplate) *batc
 	backoffLimit := int32(0)
 	jobTemplate.Spec.RestartPolicy = corev1.RestartPolicyNever
 
-	// Ensure that the job name (after suffixing with `-console`) does not exceed 57
-	// characters, to allow an additional 6 characters to appended when the job
-	// creates a pod without truncation of the `-console` suffix.
-	jobName := fmt.Sprintf("%s-%s", truncateString(r.name.Name, 49), "console")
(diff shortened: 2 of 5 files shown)


pkg/workloads/console/controller.go
```diff
@@ -182,7 +182,14 @@ func (r *reconciler) Reconcile() (res reconcile.Result, err error) {
 		}
 	}
 
-	// Create or update the job
+	// Only create a job when the console is created. If the job is removed,
+	// return a reconciliation error.
+	if !r.console.Creating() {
+		if _, err = r.getJob(); err != nil {
+			return res, errors.Wrap(err, "failed to retrieve console job")
```
benwh, line 189: It feels a bit odd to be returning a reconciliation error here, rather than continuing but logging a warning event, or updating the console's status (phase `Destroyed`?)

I think this may have a negative side-effect of firing alerts, because we have a metric for reconcile failures.

dyson: Yeah I was 50/50 on this and wondering what people might think. 

My reason for the error is that deleting the job out from under the console is unexpected and shouldn't happen under normal operation. If we accept that statement as true, then I think we should error, fire alerts, be painful and attention seeking, when this happens. 

But I can see the flip side, where we say deleting a job is a valid way to kill a console, in which case we should mark the console as stopped (or more accurately like you mention, introduce a new phase called destroyed).

Having written this out now I think I have convinced myself of the latter and agree with you. We get to choose what is normal and supported behaviour and I can't see a downside to supporting the deletion of jobs.

I'll update this to introduce a destroyed phase which will be set if the job is missing with the same behaviour as the stopped phase - requeue after the TTL and and GC the console.

pkg/workloads/console/acceptance/acceptance.go
```diff
@@ -175,6 +175,56 @@ func (r *Runner) Run(logger kitlog.Logger, config *rest.Config) {
...
+			err = client.BatchV1().Jobs(namespace).Delete(jobName, &metav1.DeleteOptions{PropagationPolicy: &policy})
+			Expect(err).NotTo(HaveOccurred(), "could not delete console job")
+
+			By("Expect that the job no longer exists")
+			Eventually(func() error {
+				_, err = client.BatchV1().Jobs(namespace).Get(consoleName, metav1.GetOptions{})
+				return err
+			}).Should(HaveOccurred(), "expected not to find job, but did")
+
+			time.Sleep(time.Second)
```
benwh, line 219: This seems a bit racy. Could we instead do a check that *eventually* the console's status changes to stopped (or destroyed)?

dyson: Yeah this is definitely racey and we can now check with eventually that its status is updated to destroyed like you mention.

dyson:
This needs a new full review as lots has changed since the original PR. 

We now mark consoles as destroyed if the job has been deleted and perform GC on them the same as if the console had been stopped.

Also while doing this work I noticed we watch on all console events. This meant that when we updated a consoles status, we would trigger another reconciliation. I have changed this to only watch for console creations. All other reconciliations should be triggered by events happening on the child objects we watch.

Finally, in allowing for jobs to be deleted, more parts of the code needed to be changed compared to the original implementation as were no longer exiting reconciliation early with an error -  reconciliation now needs to fully support a console with no job.

pkg/workloads/console/controller.go
```diff
@@ -347,21 +383,30 @@ func (r *reconciler) clampTimeout(template *workloadsv1alpha1.ConsoleTemplate) e
...
-		InNamespace(r.name.Namespace).
-		MatchingLabels(map[string]string{"job-name": job.ObjectMeta.Name})
-	if err := r.client.List(r.ctx, opts, podList); err != nil {
-		return err
-	}
-	if len(podList.Items) > 0 {
-		pod = &podList.Items[0]
+	var newStatus workloadsv1alpha1.ConsoleStatus
+
+	if job == nil {
```
benwh, line 388: Putting a conditional here seems a bit wrong. Can we not instead conditionally wrap the bit of the code that does the pod listing so that it doesn't attempt to run if the job is nil.

Then we can keep doing all of the status calculation stuff in the `calculateStatus` func?

pkg/workloads/console/controller.go
```diff
@@ -388,6 +433,10 @@ func (r *reconciler) updateStatus(job *batchv1.Job) error {
 		auditLog(r.logger, r.console, &duration, ConsoleEnded, "Console ended after reach expiration time")
 	}
 
+	if newStatus.Phase == workloadsv1alpha1.ConsoleDestroyed {
```
benwh, line 436: The fact that we're always logging when the phase is destroyed, rather than only when we transition to destroyed, seems odd? Is this related to why we've got the predicates added?

pkg/workloads/console/controller.go
benwh: Yeah after looking at the changes in the `calculateStatus` func below, I do think it'd be nicer to keep this logic in the `calculatePhase` function

pkg/workloads/console/controller.go
```diff
@@ -102,7 +117,9 @@ func Add(ctx context.Context, logger kitlog.Logger, mgr manager.Manager, opts ..
 	}
 
 	err = ctrl.Watch(
-		&source.Kind{Type: &workloadsv1alpha1.Console{}}, &handler.EnqueueRequestForObject{},
+		&source.Kind{Type: &workloadsv1alpha1.Console{}},
+		&handler.EnqueueRequestForObject{},
+		CreatePredicate{},
```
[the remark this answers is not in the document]
dyson: Spoke about this and going to revert the change. Essentially calling the extra reconciles on update are harmless and we want to support manual updates as we could, although in practice we wouldn't, edit a console. If we don't listen to update events, our console could end up out os sync.

pkg/workloads/console/controller.go
```diff
@@ -182,10 +199,20 @@ func (r *reconciler) Reconcile() (res reconcile.Result, err error) {
...
-		return res, err
+	var job *batchv1.Job
+
+	// Try to get the consoles job
+	if j, err := r.getJob(); err == nil {
+		job = j
+	}
+
+	// Only create/update a job when the console is creating or when one
+	// already exists
```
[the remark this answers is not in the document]
dyson: Yep! Much better!

pkg/workloads/console/controller.go
```diff
@@ -213,7 +239,7 @@ func (r *reconciler) Reconcile() (res reconcile.Result, err error) {
 			return res, err
 		}
 
-	case r.console.Stopped():
+	case r.console.Stopped() || r.console.Destroyed():
```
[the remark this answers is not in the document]
dyson: Agreed on adding an `Active()` helper and using it in the two places.

pkg/workloads/console/controller.go
dyson: Agreed about this. It was wrong to set a phase outside of `calculatePhase`. I'll rejig this so we do it in there.

pkg/workloads/console/controller.go
dyson: Ah yeah good catch! I'll change the conditional to only log when the existing state isn't already `workloadsv1alpha1.ConsoleDestroyed`.

PR #146, opened 2020-04-16
Refactor consoles create to separate the functionality from flags

jackatbancast:
This change separates the ``create`` command from the flags that drive it.
This allows for easier integration within other projects, and sets a standard for other commands to be added.

### DIFF
--- a/cmd/theatre-consoles/main.go
+++ b/cmd/theatre-consoles/main.go
@@ -20,20 +20,27 @@ import (
 )
 
 var (
-	cli = kingpin.New("consoles", "Manages theatre consoles")
-
+	cli        = kingpin.New("consoles", "Manages theatre consoles")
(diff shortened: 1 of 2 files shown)


cmd/theatre-consoles/main.go
```diff
@@ -87,7 +94,13 @@ func Run(ctx context.Context, logger kitlog.Logger) error {
 	// Match on the kingpin command and enter the main command
 	switch cmd {
 	case create.FullCommand():
-		return Create(ctx, logger, runner, namespace)
+		return Create(
+			ctx, logger, runner,
+			CreateOptions{
+				Namespace: namespace,
```
dyson, line 100: Just use *cliNamespace here?

dyson: Given that if we inline the newRunner() we probably wont have the assignment lines anymore

```
// Get context and namespace from flags
kctx := *cliContext
namespace := *cliNamespace
```

PR #154, opened 2020-04-20
Add golangci-lint to CI

jackatbancast:
This change introduces the GolangCI linter into the CI, alongside
some sensible defaults for enabled linters.
This should help ensure that we are producing high quality code,
whilst raising potential issues

It is important to note that these configurations are not set in stone,
and are likely to change to suit the workflow of the project, enabling or
disabling behaviour as seen fit.


.golangci.yml
```diff
@@ -0,0 +1,207 @@
...
+  # Show only new issues: if there are unstaged changes or untracked files,
+  # only those changes are analyzed, else only changes in HEAD~ are analyzed.
+  # It's a super-useful option for integration of golangci-lint into existing
+  # large codebase. It's not practical to fix all existing issues at the moment
+  # of integration: much better don't allow issues in new code.
+  # Default is false.
+  new: false
+
+  # Show only new issues created after git revision `origin/master`
+  new-from-rev: origin/master
```
benwh, line 207: I already use `golangci-lint` locally and would prefer for this to not be set locally. Can we use to the `--new-from-rev` CLI flag instead?

It'd be really nice if we could target the base branch rather than just `origin/master` too, but apparently CircleCI don't give you an env var for that, there's various hacky solutions which people have come up with though...

jackatbancast: I thought about this too. I decided to specify it in the config as this would unify the local run with what is in CI, but happy to specify this as a flag in the step instead.

benwh: For me personally it's confusing to not be seeing all linting errors across the project while working in my editor, and feels like it's less likely we'd ever go and clean the project up if this was the case. Happy to get another opinion though.

ttamimi:
Closing stale PR

PR #159, opened 2020-04-21
Reconcile consoles with authorisation

dyson:
commit b2a4b7947d50bb6f33e4d23b6b2f62c9cf0e9b93
    Remove owner from console authorisation

    Remove the owner from the console authorisation and instead look up the
    console via the consoleRef to get the console user.

    It was thought that having the owner of the console as a field in the
    console authorisation would simplify the implementation and while it did to
    a small degree it made the console authorisations difficult to test.

    Kind only previsions a single user when creating the cluster and we are
    unable to provision more. This prevents us from being able to submit an
    authorisation as another user - users can only add themselves to the
    authorisations and a user cannot authorise their own console.

    We do support service accounts as authorisation subjects but
    unfortunately impersonation in the go client doesn't support service
    accounts and so we can't impersonate one when submitting the
    authorisation.

    As console authorisations require updating from end users, the owner
    fields immutability was enforced on the console authorisations
    validating webhook. This left us with no way to modify it.

    Removing the owner field and looking up the user from the consoleRef
    allows us to test the authorisation flow and is arguably an improvement,
    as now we don't have state split over two objects that could get out of
    sync.

    As our implementation of consoles requires that users don't have
    permission to update consoles, we can trust that the user in the console
    is correct. And as we don't lock down consoles in the kind test cluster,
    we can update the consoles users to another user before performing a
    validation.

commit 8542729bd063023af6dedcf99642174e3ac66409
    Add console authorisation phases to reconcile

    Adds the concept of authorisation to the console life-cycle.

    If a console template has no rules, then a console is automatically
    authorised and it functions the same as it did before this change.

    If a console template has rules, it now waits for authorisation with a
    `Pending Authorisation` phase. On each update the to the console
    authorisation the console will reconcile again and it will check if the
    number of authorisations matches the required number of authorisations
    specified in the rule the command matches (or the default rule). When
    the number has been met, the console console is allowed to continue to
    `Pending` as normal.

---

Two areas for improvement that have been left out that I think are best to be added after this PR is merged are:
- Increased acceptance test coverage (only testing the happy path at the moment, but really should test unhappy paths to ensure the console is secure)
- Consoles `Pending Authorisation` won't ever be garbage collected. Some discussion around how we might best want to do this would be good. I think there are lots of options for this.

### DIFF
--- a/pkg/apis/workloads/v1alpha1/helpers.go
+++ b/pkg/apis/workloads/v1alpha1/helpers.go
@@ -12,6 +12,17 @@ func (c *Console) Creating() bool {
 	return c.Status.Phase == ""
 }
 
+// PendingAuthorisation returns true if the is Pending Authorisation
+func (c *Console) PendingAuthorisation() bool {
+	return c.Status.Phase == ConsolePendingAuthorisation
+}
+
+// PendingJob returns true if the console is in a phase that occurs before job
+// creation
+func (c *Console) PendingJob() bool {
+	return c.Creating() || c.PendingAuthorisation()
+}
+
 // Pending returns true if the console is Pending
 func (c *Console) Pending() bool {
 	return c.Status.Phase == ConsolePending
--- a/pkg/workloads/console/acceptance/acceptance.go
+++ b/pkg/workloads/console/acceptance/acceptance.go
@@ -21,6 +21,7 @@ const (
 	consoleName  = "console-0"
 	templateName = "console-template-0"
 	jobName      = "console-0-console"
+	user         = "kubernetes-admin"
 )
 
 // This clientset is a union of the default kubernetes clientset and the workloads client.
@@ -78,12 +79,13 @@ func (r *Runner) Run(logger kitlog.Logger, config *rest.Config) {
 		Specify("Happy path", func() {
 			By("Create a console template")
 			var ttl int32 = 10
-			template := buildConsoleTemplate(&ttl, false)
+			template := buildConsoleTemplate(&ttl, true)
 			template, err := client.WorkloadsV1Alpha1().ConsoleTemplates(namespace).Create(template)
 			Expect(err).NotTo(HaveOccurred(), "could not create console template")
 
 			By("Create a console")
 			console := buildConsole()
+			console.Spec.Command = []string{"sleep", "666"}
 			console, err = client.WorkloadsV1Alpha1().Consoles(namespace).Create(console)
 			Expect(err).NotTo(HaveOccurred(), "could not create console")
 
@@ -99,6 +101,43 @@ func (r *Runner) Run(logger kitlog.Logger, config *rest.Config) {
 					return err
 				}).Should(HaveOccurred(), "expected console template to be deleted, it still exists")
 			}()
+
+			By("Expect an authorisation has been created")
+			Eventually(func() error {
+				_, err = client.WorkloadsV1Alpha1().ConsoleAuthorisations(namespace).Get(consoleName, metav1.GetOptions{})
+				return err
+			}).ShouldNot(HaveOccurred(), "could not find authorisation")
+
+			By("Expect the console phase is pending authorisation")
+			Eventually(func() workloadsv1alpha1.ConsolePhase {
+				console, err = client.WorkloadsV1Alpha1().Consoles(namespace).Get(consoleName, metav1.GetOptions{})
+				Expect(err).NotTo(HaveOccurred(), "could not find console")
+				return console.Status.Phase
+			}).Should(Equal(workloadsv1alpha1.ConsolePendingAuthorisation))
+
+			By("Expect that the job has not been created")
+			Eventually(func() error {
+				_, err = client.BatchV1().Jobs(namespace).Get(consoleName, metav1.GetOptions{})
+				return err
+			}).Should(HaveOccurred(), "expected not to find job, but did")
+
+			By("Expect a console authorisation has been created")
+			consoleAuthorisation := &workloadsv1alpha1.ConsoleAuthorisation{}
+			Eventually(func() error {
+				consoleAuthorisation, err = client.WorkloadsV1Alpha1().ConsoleAuthorisations(namespace).Get(consoleName, metav1.GetOptions{})
+				return err
+			}).ShouldNot(HaveOccurred(), "could not find console authorisation")
+
+			// Change the console user to another user as a user cannot authorise their own console
+			By("Update the console user")
+			console.Spec.User = "another-user@example.com"
+			console, err = client.WorkloadsV1Alpha1().Consoles(namespace).Update(console)
+			Expect(err).NotTo(HaveOccurred(), "could not update console user")
+
+			By("Authorise a console")
+			consoleAuthorisation.Spec.Authorisations = []rbacv1.Subject{{Kind: "User", Name: user}}
+			_, err = client.WorkloadsV1Alpha1().ConsoleAuthorisations(namespace).Update(consoleAuthorisation)
+			Expect(err).NotTo(HaveOccurred(), "could not authorise console")
 
 			By("Expect a job has been created")
 			Eventually(func() error {
@@ -125,7 +164,7 @@ func (r *Runner) Run(logger kitlog.Logger, config *rest.Config) {
 				console, err = client.WorkloadsV1Alpha1().Consoles(namespace).Get(consoleName, metav1.GetOptions{})
 				Expect(err).NotTo(HaveOccurred(), "could not find console")
 				return console.Status.Phase
-			}, 8*time.Second).Should(Equal(workloadsv1alpha1.ConsoleStopped))
+			}).Should(Equal(workloadsv1alpha1.ConsoleStopped))
 
 			// TODO: attach to pod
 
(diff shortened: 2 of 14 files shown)


vendor/github.com/onsi/gomega/go.mod
```diff
@@ -1,5 +1,7 @@
 module github.com/onsi/gomega
 
+go 1.14
```
dyson, line 3: Hmm... really? No idea how that would have gotten in there. The tests all pass so figuring this is ok...

config/overlays/acceptance/console-authoriser.yaml
```diff
@@ -0,0 +1,7 @@
+---
```
dyson, line 1: Hmm this shouldn't have been added. Deleting!

config/overlays/acceptance/kustomization.yaml
```diff
@@ -6,6 +6,7 @@ bases:
   - ../../base
 
 resources:
+  - console-authoriser.yaml
```
dyson, line 9: Also needs the ✂️

pkg/workloads/console/controller.go
```diff
@@ -373,7 +409,24 @@ func (r *reconciler) clampTimeout(template *workloadsv1alpha1.ConsoleTemplate) e
...
+	if rule == nil {
+		return true, nil
+	}
+
+	auth, err := r.getConsoleAuthorisation()
+	if err != nil {
+		return false, err
+	}
+
+	if len(auth.Spec.Authorisations) == rule.ConsoleAuthorisers.AuthorisationsRequired {
```
benwh, line 422: To protect against two people simultaneously authorising; or someone adding an additional authorisation after the console has been created?

```suggestion
	if len(auth.Spec.Authorisations) >= rule.ConsoleAuthorisers.AuthorisationsRequired {
```

pkg/workloads/console/controller.go
```diff
@@ -457,12 +516,16 @@ func calculateStatus(csl *workloadsv1alpha1.Console, job *batchv1.Job, pod *core
...
 
-	newStatus.Phase = calculatePhase(job, pod)
+	newStatus.Phase = calculatePhase(authorised, job, pod)
 
 	return newStatus
 }
 
-func calculatePhase(job *batchv1.Job, pod *corev1.Pod) workloadsv1alpha1.ConsolePhase {
+func calculatePhase(authorised bool, job *batchv1.Job, pod *corev1.Pod) workloadsv1alpha1.ConsolePhase {
+	if !authorised {
```
benwh, line 525: I reckon this block should go down to just before where we return `ConsolePending`. Assuming no programming errors it's maybe irrelevant, but regardless I think that all of the other conditions that we check for should take precedence before this one?

pkg/workloads/console/controller.go
```diff
@@ -619,40 +682,15 @@ func buildDirectoryRoleBinding(name types.NamespacedName, role *rbacv1.Role, sub
...
-	// Firstly evaluate the console command against the authorisation rules in
-	// the template. This *could* perpetually fail, if the console template is
-	// setup incorrectly, therefore by doing this before any other creations we
-	// won't add any objects to the cluster unless we know they can be used.
-	rule, err := template.GetAuthorisationRuleForCommand(command)
-	if err != nil {
-		return errors.Wrap(err, "failed to determine authorisation rule for console command")
-	}
-
+func (r *reconciler) createAuthorisationObjects(subjects []rbacv1.Subject) error {
```
benwh, line 685: This function is looking much neater now, nice!

vendor/github.com/onsi/gomega/go.mod
benwh: The fact that only this `go.mod` file, out of all the ones in `vendor/`, has been updated makes me think that your editor has done something weird. Revert this probably?

pkg/workloads/console/acceptance/acceptance.go
```diff
@@ -57,23 +58,18 @@ func (r *Runner) Prepare(logger kitlog.Logger, config *rest.Config) error {
 
 func (r *Runner) Run(logger kitlog.Logger, config *rest.Config) {
 	Describe("Consoles", func() {
-		var client clientset
-
-		BeforeEach(func() {
```
benwh, line 62: Why make this change? Was it to improve performance?

I don't think we should be just dropping the `BeforeEach` block.

As per the Ginkgo docs:
> In general, the only code within a container block should be an It block or a BeforeEach/JustBeforeEach/JustAfterEach/AfterEach block, or closure variable declarations. It is generally a mistake to make an assertion in a container block.

Maybe we want to put this stuff in a `BeforeSuite` instead?

pkg/workloads/console/acceptance/acceptance.go
```diff
@@ -57,23 +58,18 @@ func (r *Runner) Prepare(logger kitlog.Logger, config *rest.Config) error {
...
-				if err != nil {
-					logger.Log("error", err)
-					return false
-				}
-				return true
-			}).Should(Equal(true))
-		})
+		logger.Log("msg", "starting test")
+
+		client := newClient(config)
```
benwh, line 63: It may also be detrimental to be only calling this once, rather than once per spec, as well.

IIRC the client does some caching, so we might now be seeing pollution between test cases.

pkg/workloads/console/acceptance/acceptance.go
```diff
@@ -261,12 +257,56 @@ func (r *Runner) Run(logger kitlog.Logger, config *rest.Config) {
...
+				podList, err := client.CoreV1().Pods(namespace).List(opts)
+				return podList.Items, err
+			}).Should(HaveLen(1), "expected to find a single pod")
+
+			By("Expect the console phase is Running")
+			Eventually(func() workloadsv1alpha1.ConsolePhase {
+				console, err = client.WorkloadsV1Alpha1().Consoles(namespace).Get(consoleName, metav1.GetOptions{})
+				Expect(err).NotTo(HaveOccurred(), "could not find console")
+				return console.Status.Phase
+			}).Should(Equal(workloadsv1alpha1.ConsoleRunning))
```
benwh, line 309: Arguably these two checks could be dropped, as they're duplicating what we have in the main consoles happy-path test. No harm leaving them in if they're quick to execute though.

vendor/github.com/onsi/gomega/go.mod
dyson: Yeah lets revert it 👍

pkg/apis/workloads/v1alpha1/helpers.go
```diff
@@ -9,7 +9,12 @@ import (
 
 // Creating returns true if the console has no status (the console has just been created)
```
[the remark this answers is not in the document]
dyson: 👍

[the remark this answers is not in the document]
dyson: 👍

pkg/workloads/console/acceptance/acceptance.go
dyson: Going to revert this as waiting for the mutating webhook configurations while only needed to be done once can just be done in the before each without issue to simply things as we want a new client for each test.

pkg/workloads/console/acceptance/acceptance.go
dyson: Cool will revert this just to be safe as it doesn't really matter.

pkg/workloads/console/acceptance/acceptance.go
dyson: Yeah this is true, I was just excited to see this working!

Will remove them as once the job is created we are all good!

pkg/workloads/console/controller.go
```diff
@@ -391,14 +444,20 @@ func (r *reconciler) updateStatus(job *batchv1.Job) error {
...
 	// If there's no changes in status, don't unnecessarily update the object.
 	// This would cause an infinite loop!
 	if reflect.DeepEqual(r.console.Status, newStatus) {
 		return nil
 	}
 
+	// Console phase from Pending Authorisation to Pending
+	if newStatus.Phase == workloadsv1alpha1.ConsolePending &&
+		r.console.Status.Phase == workloadsv1alpha1.ConsolePendingAuthorisation {
+		auditLog(r.logger, r.console, nil, ConsoleStarted, "Console pending")
```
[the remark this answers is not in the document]
dyson: Yeah I think this should be written the same way as the the destroyed state. So any transition out of the pending authorisation should log authorised.

pkg/workloads/console/controller.go
```diff
@@ -436,7 +495,7 @@ func (r *reconciler) updateStatus(job *batchv1.Job) error {
 	return nil
 }
 
-func calculateStatus(csl *workloadsv1alpha1.Console, job *batchv1.Job, pod *corev1.Pod) workloadsv1alpha1.ConsoleStatus {
+func calculateStatus(csl *workloadsv1alpha1.Console, authorised bool, job *batchv1.Job, pod *corev1.Pod) workloadsv1alpha1.ConsoleStatus {
```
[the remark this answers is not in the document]
dyson: Discussed briefly and can possibly see it would be nice to have along with some other info like time of authorisation on the console status, but for now going to leave as is. We can always easily add this later if we feel the need.

pkg/workloads/console/acceptance/acceptance.go
dyson: Actually going top merge this with the happy path test. Everything else is covered in integration tests.

pkg/workloads/console/acceptance/acceptance.go
```diff
@@ -125,7 +162,7 @@ func (r *Runner) Run(logger kitlog.Logger, config *rest.Config) {
 				console, err = client.WorkloadsV1Alpha1().Consoles(namespace).Get(consoleName, metav1.GetOptions{})
 				Expect(err).NotTo(HaveOccurred(), "could not find console")
 				return console.Status.Phase
-			}, 8*time.Second).Should(Equal(workloadsv1alpha1.ConsoleStopped))
```
dyson, line 128: Note sure we actually get value from this. I had a test flake locally because of this timeout. It should normally be fast but worse case we wait a minute and then fail. If we are hung here, locally I find it useful to poke around the test cluster to see what is going on. If we hang in CI waiting a little longer doesn't seem a big deal to me.

pkg/workloads/console/acceptance/acceptance.go
```diff
@@ -137,7 +174,7 @@ func (r *Runner) Run(logger kitlog.Logger, config *rest.Config) {
 			Eventually(func() error {
 				_, err = client.WorkloadsV1Alpha1().Consoles(namespace).Get(consoleName, metav1.GetOptions{})
 				return err
-			}, 12*time.Second).Should(HaveOccurred(), "expected not to find console, but did")
```
dyson, line 140: Same reasoning as above.

pkg/workloads/console/controller.go
```diff
@@ -391,17 +439,28 @@ func (r *reconciler) updateStatus(job *batchv1.Job) error {
...
+	// Console phase from Pending Authorisation
+	if newStatus.Phase != workloadsv1alpha1.ConsolePendingAuthorisation &&
+		r.console.PendingAuthorisation() {
+		auditLog(r.logger, r.console, nil, ConsoleAuthorised, "Console authorised")
+	}
+
 	// Console phase from Pending to Running
 	if newStatus.Phase == workloadsv1alpha1.ConsoleRunning &&
-		r.console.Status.Phase == workloadsv1alpha1.ConsolePending {
+		r.console.Pending() {
```
dyson, line 463: Lets use the helper functions we have! Plus I think it's easier to read.

pkg/workloads/console/acceptance/acceptance.go
benwh: Yeah let's bump this up, I think that makes sense and shouldn't impact anything else.

pkg/workloads/console/acceptance/acceptance.go
benwh: This case is a little different, as it's related to the 10-second `DefaultTTLSecondsAfterFinished` value, but I guess if we see it flake then we'll need to bump it too.

pkg/workloads/console/controller.go
benwh: Nice, agreed!

benwh: I do think we should have swapped the order of these comparisons though, I find it much easier to visualise the transition when reading `{old,new}` rather than `{new,old}` - but that's out of the scope of this PR!

pkg/workloads/console/controller.go
dyson: We aren't able to move this as the function is written but we would still like to improve this function. As it is currently written it assumes that the input arguments are correct, but if they aren't it can have odd behaviour and return the wrong phase.

pkg/workloads/console/acceptance/acceptance.go
dyson: Yeah let's add it back, I was probably a little heavy handed before removing this as well.

pkg/workloads/console/controller.go
dyson: Oh yeah that's totally right. I new something was feeling weird about the comparisons and I think this is it!

dyson: Ended up just doing this change while I had my editor open.

dyson:
Everything should be added and good for another look over.

PR #160, opened 2020-04-22
Add hooks to console phase changes in cli helpers

jackatbancast:
This change introduces hooks for phase change points in the cli
helpers.
These can be used to log when a phase change occurs, as we have done in
``theatre-consoles``.


jackatbancast:
@benwh Having a look at your suggestions it feels like adding the interface with methods is probably the best tradeoff to make. I changed some of the names where I felt appropriate, but otherwise brought in your changes directly. Thanks for sketching up the changes.

PR #162, opened 2020-04-23
Add pending approval TTL and improve GC

dyson:
With the introduction of authorisations consoles that need
authorisations enter a Pending Authorisation state. If the console is
never authorised, we need to ensure we clean them up and don't just
leave them hanging forever.

We introduce a new field on the console template (and therefore
template) to set the TTL for a console to reach the Running phase. If a
console doesn't reach this phase before the TTL (including if it is
still in the Pending Authorisation phase) then we garbage collect it.

### DIFF
--- a/pkg/apis/workloads/v1alpha1/helpers.go
+++ b/pkg/apis/workloads/v1alpha1/helpers.go
@@ -43,33 +43,60 @@ func (c *Console) Destroyed() bool {
 	return c.Status.Phase == ConsoleDestroyed
 }
 
-// Active returns true is the console is active
-func (c *Console) Active() bool {
-	return c.Creating() || c.Pending() || c.Running()
+// PreRunning returns true if the console is in a phase before Running
+func (c *Console) PreRunning() bool {
+	return c.Creating() || c.PendingAuthorisation() || c.Pending()
 }
 
-// EligibleForGC returns true if the console can be garbage collected. This is the case if
-// its TTLSecondsAfterFinished has elapsed.
+// PostRunning returns true if the console is in a phase after Running
+func (c *Console) PostRunning() bool {
+	return c.Stopped() || c.Destroyed()
+}
+
+// EligibleForGC returns whether a console can be garbage collected
 func (c *Console) EligibleForGC() bool {
-	if c.Active() {
+	gcTime := c.GetGCTime()
+	if gcTime == nil {
 		return false
 	}
 
-	if c.Status.ExpiryTime == nil {
-		return false
-	}
-
-	// When the console is completed
-	if c.Status.CompletionTime != nil {
-		return c.Status.CompletionTime.Time.Add(c.TTLDuration()).Before(time.Now())
-	}
-
-	return c.Status.ExpiryTime.Time.Add(c.TTLDuration()).Before(time.Now())
+	return gcTime.Before(time.Now())
 }
 
-// TTLDuration returns the console's TTL as a time.Duration
-func (c *Console) TTLDuration() time.Duration {
+// GetGCTime returns time time at which a console can be garbage collected, or
+// nil if it cannot be.
+//
+// This will be the case if:
+// - TTLSecondsBeforeRunning has elapsed and the console hasn't progressed to running
+// - TTLSecondsAfterFinished has elapsed and the console is stopped or destroyed
+func (c *Console) GetGCTime() *time.Time {
+	switch {
+	case c.PreRunning():
+		// When the console hasn't progressed to the running phase
+		t := c.CreationTimestamp.Add(c.TTLSecondsBeforeRunning())
+		return &t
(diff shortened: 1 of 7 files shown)


dyson:
Commit with a test for this coming - just wanted to get this up.

pkg/apis/workloads/v1alpha1/helpers.go
```diff
@@ -43,35 +43,51 @@ func (c *Console) Destroyed() bool {
 	return c.Status.Phase == ConsoleDestroyed
 }
 
-// Active returns true is the console is active
+// Active returns true if the console is active
```
dyson, line 46: This is not longer used ✂️

pkg/workloads/console/controller.go
```diff
@@ -316,21 +319,29 @@ func (r *reconciler) setConsoleOwner(consoleTemplate *workloadsv1alpha1.ConsoleT
...
+	if r.console.Spec.TTLSecondsAfterFinished != nil {
+		updatedCsl.Spec.TTLSecondsAfterFinished = r.console.Spec.TTLSecondsAfterFinished
+	} else if consoleTemplate.Spec.DefaultTTLSecondsAfterFinished != nil {
 		updatedCsl.Spec.TTLSecondsAfterFinished = consoleTemplate.Spec.DefaultTTLSecondsAfterFinished
 	} else {
-		updatedCsl.Spec.TTLSecondsAfterFinished = &defaultTTL
+		updatedCsl.Spec.TTLSecondsAfterFinished = &defaultTTLSecondsAfterFinished
 	}
 
 	if err := r.client.Update(r.ctx, updatedCsl); err != nil {
```
benwh, line 343: Because we've dropped the early return, are we now always doing updates to the console whenever we go through the reconcile loop?

pkg/workloads/console/acceptance/acceptance.go
```diff
@@ -186,9 +187,43 @@ func (r *Runner) Run(logger kitlog.Logger, config *rest.Config) {
 			}).Should(Equal(0), "pod did not get deleted")
 		})
 
+		Specify("Fails to run", func() {
```
benwh, line 190: nit: Can we change the name of this? It's a bit ambiguous without reading the test specs to understand what's going on.

Maybe:
> Automatically deleted if not authorised within TTLBeforeRunning

Or something more concise? 😆

pkg/workloads/console/controller.go
```diff
@@ -221,8 +222,12 @@ func (r *reconciler) Reconcile() (res reconcile.Result, err error) {
 	}
 
 	switch {
-	case r.console.Pending():
-		res = requeueAfterInterval(r.logger, time.Second)
+	case r.console.PreRunning():
```
benwh, line 225: Because PreRunning encapsulates `{Creating, PendingAuthorisation, Pending}`, the behaviour of this case block has now changed: We used to requeue every second if in the `Creating` phase, but now we reconcile after `TTLSecondsBeforeRunning()`.

I think maybe we should flip this, and it's more correct to have the conditional be specific to the `PendingAuthorisation` phase, else we reconcile in a second. Might be more defensible to future code changes?

pkg/workloads/console/controller.go
```diff
@@ -221,8 +222,12 @@ func (r *reconciler) Reconcile() (res reconcile.Result, err error) {
...
+		if r.console.Pending() {
+			res = requeueAfterInterval(r.logger, time.Second)
+		} else {
+			res = requeueAfterInterval(r.logger, r.console.TTLSecondsBeforeRunning())
```
benwh, line 229: This _might_ be technically wrong? (If so, it definitely was before too, but we could take this opportunity to fix it up)

I'm thinking this timeline:

Timestamp | Event
--------- | ------
0s | Create a console, with TTLSecondBeforeRunning of 300s
50s | Controller is restarted. A new watch on the object starts, and we hit this code
300s | We should have begun reconciling the console here, because we hit (CreationTimestamp + TTLSecondsBeforeRunning)
350s | We actually begin reconciling here. The console is GC'd 50s after it should have been.

pkg/workloads/console/controller.go
benwh: Discussed: We'll try to remove this first conditional, and only requeue after TTLSecondsBeforeRunning

pkg/workloads/console/controller.go
dyson: Ah yeah good catch!

pkg/workloads/console/controller.go
dyson: Ah yep this is totally true!

benwh: Discussed: We'll try replacing this with a call to CreateOrUpdate; otherwise re-add the conditional at the top and check both fields

PR #170, opened 2020-04-30
Add audit event for consoles expiring without auth

benwh:
2857899: Move TTLBeforeRunning test to integration suite

There's no need for this to be in the acceptance test suite, and while
developing a related change I found it was a much quicker feedback loop
to have it in integration, due to not having to re-prepare my cluster
for each test run after a code change.

b609c57: Add audit event for consoles expiring without auth

This will ensure that we emit a log entry with `event=ConsoleEnded`
and `console_requires_authorisation=true`, allowing us to determine if
any consoles were started, required authorisation, but were not actually
authorised.

29e7368: Add note about ns deletion in integration tests

I was under the impression that the per-spec namespaces were removed
after each test run, but this is not the case as per the comment below
the added line.

Add a log message to remind the reader of integration test logs that
this isn't happening. This will hopefully make it clearer as to why
there are log entries from different integration specs' reconciliations
interleaved with each other.

4dec51f: Fix nil pointer dereference in console logging

This has been breaking our integration tests intermittently for some
time.

The issue was in the following conditional:
```go
if newStatus.CompletionTime != r.console.Status.CompletionTime && !statusCtx.Job.Status.StartTime.IsZero() {
...
```

In the first part of the expression we were comparing two pointers
(`CompletionTime`) for inequality. As far as I can tell this is *always*
going to evaluate to true, because they're different objects, so was
potentially wrong anyway.

We then get to the second part of the expression and hit a nil pointer
dereference because the `Job` field is not populated on `statusCtx`.
This is actually valid: It will happen when the console is in the
`Destroyed` phase (the job has been deleted).

Fix this by changing the conditional: Inspect the new status' phase and
only log this event if it's equal to `Stopped`.

This leaves us open to potentially not logging this `ConsoleEnded` event
if the console has effectively transitioned through multiple phases
between a reconcile loop; for example from `Pending` to `Running` to
`Stopped`. I'm not sure whether this is actually possible in a real
deployment, but add a new block to capture this, in case it is.

### DIFF
--- a/pkg/workloads/console/controller.go
+++ b/pkg/workloads/console/controller.go
@@ -514,9 +514,11 @@ func (r *reconciler) generateStatusAndAuditEvents(statusCtx consoleStatusContext
 		logger.Log("event", ConsoleEnded, "msg", "Console ended", "duration", duration)
 	}
 
-	// Console phase from Running to Stopped without CompletionTime: the job's
-	// activeDeadlineSeconds was reached, the job was marked as failed and the
-	// pod deleted.
+	// Console phase from Running to Stopped without CompletionTime.
+	// Either:
+	// - The job's activeDeadlineSeconds was reached, and the job was marked as
+	//   failed and the pod deleted.
+	// - The pod ended with a non-zero exit code, and the job was marked as failed.
 	if r.console.Running() && newStatus.Phase == workloadsv1alpha1.ConsoleStopped &&
 		newStatus.CompletionTime == nil {
 		duration := r.console.Status.ExpiryTime.Sub(statusCtx.Job.Status.StartTime.Time).Seconds()
--- a/pkg/workloads/console/controller.go
+++ b/pkg/workloads/console/controller.go
@@ -298,6 +298,7 @@ func (r *reconciler) Reconcile() (res reconcile.Result, err error) {
 		if err = r.client.Delete(r.ctx, r.console, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil {
 			return res, err
 		}
+
 		res = reconcile.Result{Requeue: false}
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/workloads/console/controller.go
```diff
@@ -898,7 +908,7 @@ func getAuditLogger(logger kitlog.Logger, c *workloadsv1alpha1.Console, statusCt
 	}
 
 	if statusCtx.Authorisation != nil {
-		var subjectNames []string
+		subjectNames := []string{}
```
benwh, line 919: This is just so that in the log event we get `console_authorisers=[]` rather than `console_authorisers=null`

pkg/workloads/console/controller.go
```diff
@@ -505,19 +506,28 @@ func (r *reconciler) generateStatusAndAuditEvents(statusCtx consoleStatusContext
...
+	// pod deleted.
 	if r.console.Running() &&
 		newStatus.CompletionTime == nil && newStatus.Phase == workloadsv1alpha1.ConsoleStopped {
 		duration := r.console.Status.ExpiryTime.Sub(statusCtx.Job.Status.StartTime.Time).Seconds()
-		logger.Log("event", ConsoleEnded, "msg", "Console ended after reaching expiration time", "duration", duration)
+		logger.Log("event", ConsoleEnded, "msg", "Console ended due to expiration", "duration", duration)
+	}
+
+	// Console was in PendingAuthorisation phase, but is about to be deleted.
+	if r.console.PendingAuthorisation() && r.console.EligibleForGC() {
```
benwh, line 534: I put this here, rather than amending the logger in the block where we delete the console if it's eligible for GC because it became clear that these are two different cases: Here we're generating an audit log for human consumption (and therefore it makes sense for it to be co-located with the rest of the audit logs), and in the other block we're primarily creating a Kubernetes event, which happens to be via the logger.

pkg/workloads/console/integration/integration_test.go
```diff
@@ -638,7 +638,7 @@ var _ = Describe("Console", func() {
 						identifier, _ := client.ObjectKeyFromObject(csl)
 						err := mgr.GetClient().Get(context.TODO(), identifier, &workloadsv1alpha1.Console{})
 						return apierrors.ReasonForError(err)
-					}, 5*time.Second).Should(Equal(metav1.StatusReasonNotFound), "expected not to find console, but did")
+					}, 10*time.Second).Should(Equal(metav1.StatusReasonNotFound), "expected not to find console, but did")
```
benwh, line 641: Bumped this as I saw CI flake due to it when running with `-untilItFails`

benwh:
There was maybe a touch of scope creep in this PR with fixing the integration test, but it's all related code so hopefully it doesn't make it too much of a pain to review - if you look at it commit-by-commit.

PR #179, opened 2020-05-15
Fix ConsoleAuthorisation role permissions

benwh:
We need to supply the `patch` verb on the per-`consoleauthorisation`
roles, to allow `PATCH` updates on these objects, which is part of the
expected workflow.


dyson:
Are we actually using go releaser now?

benwh:
It will cut new versions yes, but won't actually have any influence on the Docker image we use, as we don't push public images yet.

Regardless that's a good point, I'll bump to v1.0.1

PR #182, opened 2020-06-08
Bump envconsul init container CPU limit

lawrencejones:
Depending on the app, the init container is taking ages to run. We want
this to burst to provide a better experience when deploying.

### DIFF
--- a/pkg/vault/envconsul/webhook.go
+++ b/pkg/vault/envconsul/webhook.go
@@ -319,10 +319,12 @@ func (i podInjector) buildInitContainer() corev1.Container {
 				ReadOnly:  false,
 			},
 		},
+		// This container is only alive for a brief moment, so make our request very small
+		// while permitting a big burst.
 		Resources: corev1.ResourceRequirements{
 			Limits: corev1.ResourceList{
 				corev1.ResourceMemory: resource.MustParse("64Mi"),
-				corev1.ResourceCPU:    resource.MustParse("50m"),
+				corev1.ResourceCPU:    resource.MustParse("1"),
 			},
 			Requests: corev1.ResourceList{
 				corev1.ResourceMemory: resource.MustParse("64Mi"),


pkg/vault/envconsul/webhook.go
```diff
@@ -319,10 +319,12 @@ func (i podInjector) buildInitContainer() corev1.Container {
 				ReadOnly:  false,
 			},
 		},
+		// This container is only alive for a brief moment, so make our request very small
+		// while permitting a big burst.
 		Resources: corev1.ResourceRequirements{
 			Limits: corev1.ResourceList{
 				corev1.ResourceMemory: resource.MustParse("64Mi"),
-				corev1.ResourceCPU:    resource.MustParse("50m"),
+				corev1.ResourceCPU:    resource.MustParse("1"),
```
benwh, line 327: Does this have the side-effect of effectively transforming *every* envconul-injected pod into the `Burstable` QoS class, if it would normally be `Guaranteed` without this init container pod?

Without looking at the docs I'm not sure whether init containers have an influence on this property or not.

lawrencejones: > The QoS (quality of service) tier of the Pod’s effective QoS tier is the QoS tier for init containers and app containers alike.

https://github.com/kubernetes/kubernetes/pull/75223

Urgh, I think it may. Let's change this to be 500m for both?

benwh: Sure that seems like a reasonable compromise.

Although as discussed separately, I still think this seems like overkill given how little this process is doing!

If you've got the appetite to do it in this PR, then it would be interesting to add a logger at the end of the `case install.FullCommand()` block, so we can see how long that actually takes via the timestamps

rnaveiras:
enconsul is gone, so we can close this PR

PR #183, opened 2020-06-15
Add sidecar container support to consoles

dyson:
commit 75b257202d314846b2562198ab07c4674b721d19

    Basic support for multiple containers in console

    Console templates by design only allow a single container to be defined.
    This restriction then relied on the default behaviour of kubernetes - if
    there is only a single container, don't require specifying the container
    in actions such as attaching or displaying logs.

    Relying on the default behaviour breaks when containers are injected into the
    console pod via webhooks. In this case kubernetes requires you to also
    provide the name of the container that you want to interact with.

    This change gets the first container that has a TTY (which should only
    be the single container defined in the console template) and passes it
    through to attach.

    Ideally we should support multiple containers in consoles and allow the
    specifying of a container to attach to from the command line as well as
    being able to set a default in the console template. This change however
    is the minimum to support multiple containers without changing current
    behaviour.

commit 802398d247b3fc22b96f247ee5bf334a580281ae

    Ignore nil error in theatre-consoles command

    Don't print and exit with an unexected error when theatre-console
    commands run successfully (err == nil).

commit 2cacf07f68e1d8bc31266b8b9f3321e3be189e85 (HEAD)

     v1.0.3


pkg/workloads/console/runner/runner.go
```diff
@@ -684,17 +680,17 @@ func rbHasSubject(rb *rbacv1.RoleBinding, subjectName string) bool {
 }
 
 // GetAttachablePod returns an attachable pod for the given console
-func (c *Runner) GetAttachablePod(csl *workloadsv1alpha1.Console) (*corev1.Pod, error) {
+func (c *Runner) GetAttachablePod(csl *workloadsv1alpha1.Console) (*corev1.Pod, string, error) {
```
jackatbancast, line 683: I think it'd be nicer to handle this by returning a struct. It'd be more compatible if we added more return args, but it'd also help keep context when reading a line in isolation.

That said, I think we should merge for now and I can adjust in my update to kubebuilder v2.

dyson: Yeah true that would probably be nicer. But happy to defer and to just get this in for now.

PR #184, opened 2020-06-29
WIP: Upgrade kubebuilder to v2.3.1

jackatbancast:
This upgrade involves upgrades to go.mod and newer versions of kubernetes libs and controller-{runtime,gen}.

Most of the remaining work is copying the controller code, regenerating the webhooks and re-adding the tests.

PR #185, opened 2020-07-20
Upgrade to v2 - rbac - 1/4

dyson:
This PR has been extracted out of the [upgrade-kubebuilder/v2](https://github.com/gocardless/theatre/pull/184) branch and represents migrating over our rbac controller as the as the first component in the upgrade.

It primarily consists of deleting vendor as we're now using go mod, deleting all non-rbac resources as we can't have the project in a half and half state, and then upgrading the rbac related code.

A number of files were modified too much for git to recognise them as having moved in a single commit and so performed a deletion and addition instead. I imagine a manual comparison against master side by side instead of viewing the diff would be more useful but in case you want to see a diff they can be seen with:

```
$ git diff master:config/base/crds/rbac_v1alpha1_directoryrolebinding.yaml config/base/crds/rbac.crd.gocardless.com_directoryrolebindings.yaml
$ git diff master:pkg/apis/rbac/v1alpha1/directory_role_binding_types.go apis/rbac/v1alpha1/directoryrolebinding_types.go
$ git diff master:pkg/rbac/directoryrolebinding/controller.go controllers/rbac/directoryrolebinding_controller.go
$ git diff master:pkg/rbac/directoryrolebinding/suite_test.go controllers/rbac/suite_test.go```
$ git diff master:pkg/rbac/directoryrolebinding/integration/suite_test.go controllers/rbac/integration/suite_test.go
```

This PR should be viewed in isolation as other components are to follow in subsequent individual PRs until we have everything moved over to v2.

0a47114138dd97c01aa82e2ea8b7bd78114dddaa

    Remove vendor

99b1ecd3fd16fcbea7233c4510dcb80a79a7aaa2

    Remove all non-rbac resources

    The upgrade to kubebuilder v2, along with the upgrade of
    controller-runtime and controller-gen is extremely invasive and cannot
    easily be done on top of our existing project.

    In order to make the initial change to the new structure, we have chosen
    to isolate the RBAC controller. As we can't leave half of the old v1
    project around, we need to remove all of the non-rbac related resources.

    This should allow us to create a nice diff for rbac in the next commit
    where we move rbac to v2.

    This is a best effort approach to moving to v2 with minimal pain. We
    will add these deleted files back for each resource as we add them to
    maintain the diffs.

ac751922d19d3f0d938462d044e8c1130a2c2808 (HEAD -> v2, origin/v2)

    Migrate rbac controller to kubebuilder v2

    As rbac is one of the simplest components, as well as being a dependency
    of other components in theatre, we have chosen to migrate it to v2
    first.

    This commit also includes the surrounding changes to things such as
    docker, circleci, Makefile, etc, which make up the project and are
    necessary.


apis/rbac/v1alpha1/directoryrolebinding_types.go
```diff
@@ -0,0 +1,41 @@
...
+type DirectoryRoleBindingSpec struct {
+	Subjects []rbacv1.Subject `json:"subjects"`
+	RoleRef  rbacv1.RoleRef   `json:"roleRef"`
+}
+
+// DirectoryRoleBindingStatus defines the observed state of DirectoryRoleBinding
+type DirectoryRoleBindingStatus struct{}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
```
jackatbancast, line 18: This snuck in?

cmd/helpers.go
```diff
@@ -39,38 +36,29 @@ func (opt *commonOptions) WithMetrics(cmd *kingpin.Application) *commonOptions {
...
 	// @template: caller is optional, but can be useful to highlight where the log-line is
 	// originating from. This configuration has to happen after we've leveled the logger, to
 	// avoid the caller being set to level.go:63 all the time.
-	logger = kitlog.With(logger, "ts", kitlog.DefaultTimestampUTC, "caller", logging.RecorderAwareCaller())
-
-	// Connect the Go standard library logger into our kitlog instance so other
-	// dependencies output via kitlog.
-	stdlog.SetOutput(kitlog.NewStdlibAdapter(logger))
-	klog.SetOutput(kitlog.NewStdlibAdapter(logger))
+	// logger = kitlog.With(logger, "ts", kitlog.DefaultTimestampUTC, "caller", logging.RecorderAwareCaller())
```
jackatbancast, line 57: We need to make sure we add this back in

config/base/managers/rbac.yaml
```diff
@@ -80,6 +91,7 @@ spec:
           args:
             - --google  # enable GoogleGroup resolution
             - --metrics-address=0.0.0.0
+            - --manager-leader-election
```
jackatbancast, line 94: This will probably break, as we don't have this flag in the

apis/rbac/v1alpha1/directoryrolebinding_types.go
dyson: Removed 👍

cmd/helpers.go
dyson: Added this functionality back in! 👍

Dockerfile
```diff
@@ -1,18 +1,12 @@
...
-      && make linux/amd64 \
-      && mkdir ../bin \
-      && mv pkg/linux_amd64/envconsul ../bin
+COPY go.mod go.sum /go/src/github.com/gocardless/theatre/
+RUN go mod download -x
 
 COPY . /go/src/github.com/gocardless/theatre
-RUN make VERSION=$(cat VERSION) build
+
+RUN make -j 3 VERSION=$(cat VERSION) build
```
lawrencejones, line 9: Made a comment in the other PR- we shouldn't do this, as Go already fires up all the computer has, and sequentially calling it will cache the modules for subsequent builds.

Makefile
```diff
@@ -13,53 +19,48 @@ build-all: build-darwin build-linux
 
 # Specific linux build target, making it easy to work with the docker acceptance
 # tests on OSX
-bin/%.linux_amd64:
+bin/%.linux_amd64: vet
```
lawrencejones, line 22: I'd really prefer if we don't force people to run vet in their dev flows- our preference in the past has been to enforce things in CI, but allow people to opt into linting locally.

Is there any reason we'd want to change?

README.md
```diff
@@ -1,170 +0,0 @@
-# Theatre [![CircleCI](https://circleci.com/gh/gocardless/theatre.svg?style=svg)](https://circleci.com/gh/gocardless/theatre)
```
lawrencejones, line 1: Are we meant to delete the readme?

dyson: We were planning on adding them all back at the end and ensuring they are all up to date and contain everything we want.

Makefile
dyson: Only that vet can detect errors the compiler can't. This is also a kubebuilder v2 change (as in this is what they do in their generated Makefile. We don't need to follow practice and we could add it as a CI step.

Dockerfile
dyson: Thought I commented in the other one. Increasing the jobs makes a massive difference to me when building all the binaries. I actually call it with more jobs when I am building locally but we thought 3 for the dockerfile was a reasonable compromise. I would be surprised if you didn't also find it faster.

Makefile
lawrencejones: So running vet locally is a great idea, but enforcing that you run vet on every compilation loop means you apply your opinions to another person's setup, making the choice for them.

Running it in CI means we don't merge anything that vet would fail, because you'll go red on your PR. Anyone who wants to can run vet locally before they commit, so that's easy. And everyone who just wants to build the binary is able to build the binary without also running vet, which is good.

Dockerfile
lawrencejones: You're right about subsequent builds, once your cache is populated. But with a fresh first build and no module cache, starting multiple binary targets building concurrently means we waste resources, and get the job done slower.

This is non-intuitive, especially when you've seen it make things much faster on your local machine. But watch what happens if I build the dockerfile with `time make`, increasing the concurrency (`-j`):

### `-j 1`, 47.8s

```console
Step 6/6 : RUN time make VERSION=$(cat VERSION) build                                                                                                                     
 ---> Running in 73203fc07647                                                                                                                                             
CGO_ENABLED=0 GOARCH=amd64 go build -ldflags "-s -w -X main.Version=1.0.2" -o bin/rbac-manager ./cmd/rbac-manager/.                                                       
CGO_ENABLED=0 GOARCH=amd64 go build -ldflags "-s -w -X main.Version=1.0.2" -o bin/workloads-manager ./cmd/workloads-manager/.                                             
CGO_ENABLED=0 GOARCH=amd64 go build -ldflags "-s -w -X main.Version=1.0.2" -o bin/vault-manager ./cmd/vault-manager/.                                                     
CGO_ENABLED=0 GOARCH=amd64 go build -ldflags "-s -w -X main.Version=1.0.2" -o bin/theatre-envconsul ./cmd/theatre-envconsul/.                                             
CGO_ENABLED=0 GOARCH=amd64 go build -ldflags "-s -w -X main.Version=1.0.2" -o bin/theatre-consoles ./cmd/theatre-consoles/.                                               
126.76user 139.77system 0:47.82elapsed 557%CPU (0avgtext+0avgdata 583184maxresident)k                                                                                     
6928inputs+1215272outputs (13major+3039960minor)pagefaults 0swaps  
```

### `-j 2`, 1m 5s

```
Step 6/6 : RUN time make -j 2 VERSION=$(cat VERSION) build
 ---> Running in ec69e68792dc
CGO_ENABLED=0 GOARCH=amd64 go build -ldflags "-s -w -X main.Version=1.0.2" -o bin/rbac-manager ./cmd/rbac-manager/.
CGO_ENABLED=0 GOARCH=amd64 go build -ldflags "-s -w -X main.Version=1.0.2" -o bin/workloads-manager ./cmd/workloads-manager/.
CGO_ENABLED=0 GOARCH=amd64 go build -ldflags "-s -w -X main.Version=1.0.2" -o bin/vault-manager ./cmd/vault-manager/.
CGO_ENABLED=0 GOARCH=amd64 go build -ldflags "-s -w -X main.Version=1.0.2" -o bin/theatre-envconsul ./cmd/theatre-envconsul/.
CGO_ENABLED=0 GOARCH=amd64 go build -ldflags "-s -w -X main.Version=1.0.2" -o bin/theatre-consoles ./cmd/theatre-consoles/.
205.06user 215.02system 1:05.57elapsed 640%CPU (0avgtext+0avgdata 551624maxresident)k 
0inputs+1427232outputs (5major+4390003minor)pagefaults 0swaps
```

### `-j 3`, 1m 37s

```
Step 6/6 : RUN time make -j 3 VERSION=$(cat VERSION) build
 ---> Running in 2b84d4328f0b
CGO_ENABLED=0 GOARCH=amd64 go build -ldflags "-s -w -X main.Version=1.0.2" -o bin/rbac-manager ./cmd/rbac-manager/.
CGO_ENABLED=0 GOARCH=amd64 go build -ldflags "-s -w -X main.Version=1.0.2" -o bin/workloads-manager ./cmd/workloads-manager/.
CGO_ENABLED=0 GOARCH=amd64 go build -ldflags "-s -w -X main.Version=1.0.2" -o bin/vault-manager ./cmd/vault-manager/.
CGO_ENABLED=0 GOARCH=amd64 go build -ldflags "-s -w -X main.Version=1.0.2" -o bin/theatre-envconsul ./cmd/theatre-envconsul/.
CGO_ENABLED=0 GOARCH=amd64 go build -ldflags "-s -w -X main.Version=1.0.2" -o bin/theatre-consoles ./cmd/theatre-consoles/.
306.67user 331.53system 1:37.39elapsed 655%CPU (0avgtext+0avgdata 582716maxresident)k
61872inputs+1648720outputs (327major+5836035minor)pagefaults 0swaps
```

Hopefully this shows that while it might be great for your dev machine, it won't help our docker setup. Probably the worst bit though, more than this making things generally slower, is that you genuinely do more work as the concurrency increases, as you recompile those vendored modules many times over. So our shared container builder instance will get sad.

dyson: Ah interesting. Thanks for the detailed response!

Makefile
dyson: I will add a vet step to CI instead 👍

dyson:
Going to be closing this in favour of https://github.com/gocardless/theatre/pull/188/. This new PR should be much easier to review. 

I will fix the issues addressed in here in that PR.

PR #187, opened 2020-07-21
v2: Disable make targets and ci jobs

dyson:
Disables a bunch of make targets and circle ci steps so no code is built,
tested, or generated.

As we have deleted vendor we are in a broken state and we only want to
add these back as we add v2 features (ignoring the as yet to be migrated
v1 files). This way as we move over each component we can move over
fully working tests, code generation, and ci steps.


dyson:
Yep me too - they should hopefully be very short lived and we should effectively end back at the point where they are fully commented out and being used again at the end of this process. I just didn't want to freak people out by deleting everything!

PR #192, opened 2020-07-22
Remove error recording from the Info handler

jackatbancast:
Remove error recording from the Info handler
18a35bf

We have removed all of the calls to logger.Info(...) that contained an
error. This change removes the code inside of Info that handled the
cases where it wasn't set.

Going forward, all errors should be reported using

    logger.Error(err, "some error message", ...)

and other informational messages should be recorded using

    logger.Info("some message", ...)

style.

---

 Use the provided message instaid of the error string
ff324f7

This change changes the behaviour of the recorder to use a provided
error message instead of the string representing the error. In most
cases this is currently the same as the message is passed in as
err.Error(), though this is a choice that should be made in the code,
not by the recording library.

### DIFF
--- a/pkg/logging/recorded.go
+++ b/pkg/logging/recorded.go
@@ -114,7 +114,7 @@ func (erl *EventRecordingLogger) Info(msg string, keyvals ...interface{}) {
 		keyvals = keyvals[2:]
 	}
 
-	var event, eventType, message string
+	var event string
 	var ok bool
 
 	if event, ok = kvs["event"]; !ok {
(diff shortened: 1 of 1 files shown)


pkg/logging/recorded.go
```diff
@@ -96,7 +96,7 @@ func (erl *EventRecordingLogger) Error(err error, msg string, keyvals ...interfa
 		return // don't record this event
 	}
 
-	erl.recorder.Event(erl.object, corev1.EventTypeWarning, event, err.Error())
+	erl.recorder.Event(erl.object, corev1.EventTypeWarning, event, msg)
 }
 
 // WithRecorder decorates a kitlog.Logger so that any log entries that contain an
```
dyson, line 102: Comment needs updating.

PR #194, opened 2020-07-24
Move the priority webhook into v2 structure

jackatbancast:
### Move the priority webhook into v2 structure
ec24351

This change brings the v2 webhook into the same directory structure as
the other webhooks.

### Convert the webhook to kubebuilder v2 format
123c221

This change includes only the changes needed for the webhook to be
created in the new kubebuilder format.

### Add webhook definition to config yamls
7a8c5d6

This change includes an additional entry to the config yaml manifests.
This is required for our deployment to be recognised, in both the
deployed cluster, in addition to the envtest integration tests.

### Fixup priority webhook integration tests
f61959b

This change includes the changes required for the priority webhook
integration tests.
The integration package in pkg is no longer used, and the controller
and manager are created in the suite test file.


apis/workloads/v1alpha1/priority_webhook.go
```diff
@@ -115,39 +91,45 @@ func (i *injector) Handle(ctx context.Context, req types.Request) (resp types.Re
...
 	mutateTotal.With(labels).Inc() // we are committed to mutating this pod now
 
-	logger.Log("event", "pod.assign_priority_class", "class", priorityClassName)
+	logger.Info(fmt.Sprintf("pod assigned priority class %s", priorityClassName), "event", "pod.assign_priority_class", "class", priorityClassName)
 	copy := pod.DeepCopy()
 	copy.Spec.PriorityClassName = priorityClassName
 	copy.Spec.Priority = nil
 
-	return admission.PatchResponse(pod, copy)
+	// TODO(jackatbancast): convert using JSON patch operations
```
theobarberbany, line 128: ?

jackatbancast: Patching here would probably be better than creating both objects, marshalling into JSON and then having controller runtime figure out the patches for us.

Using JSON patches directly would probably help us to structure our code in a form that we can more easily reason about.

PR #211, opened 2020-08-20
Introduce noninteractive consoles

tragiclifestories:


pkg/workloads/console/runner/runner.go
```diff
@@ -56,7 +56,7 @@ type Options struct {
 	// expect a TTY to not be present?
 	// However with these types of consoles it will not be possible to send input to them
```
Sinjo, line 57: I guess we lose this comment if we merge this! Maybe add one about why we have this flag.

controllers/workloads/console/integration/integration_test.go
```diff
@@ -241,15 +241,44 @@ var _ = Describe("Console", func() {
...
 			)
 			Expect(
-				job.Spec.Template.Spec.Containers[0].TTY).To(BeTrue(),
-				"job's first container should have tty true",
+				job.Spec.Template.Spec.Containers[0].TTY).To(BeFalse(),
+				"job's first container should have tty false",
 			)
 		})
 
+		Context("with Interactive = true", func() {
```
Sinjo, line 253: Mulling over whether this should be true by default or not. Arguably it doesn't matter too much as the consumer of the library can make the decision to opt-in to TTYs (we can certainly do that in our internal tooling that needs them enabled).

Trying to figure out which way round is less of a surprise as a user of the library.

Sinjo (commented):
From what I can see this looks good, though I'll wait for someone for Core Infra to say it does too!

I figure we should add a CLI flag here while we're at it?

cmd/theatre-consoles/main.go
```diff
@@ -26,7 +26,6 @@ var (
 	cli = kingpin.New("consoles", "Manages theatre consoles").Version(cmd.VersionStanza())
 
 	cliContext = cli.Flag("context", "Kubernetes context to target. If not provided defaults to current context").
-			Short('c').
```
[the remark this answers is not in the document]
tragiclifestories: I think this was the result of some Vim stupidity on my part ...

controllers/workloads/console/integration/integration_test.go
tragiclifestories: I think that's the caller's decision - anu consoles create defaults it to true in our test branch, FWIW. Utopia should do the same. If we do more automation in this style, however, we could potentially have more places than Dispatcher where `false` made sense.

pkg/workloads/console/runner/runner.go
```diff
@@ -232,11 +228,36 @@ func (c *Runner) Create(ctx context.Context, opts CreateOptions) (*workloadsv1al
...
 	return csl, nil
 }
 
+func (c *Runner) exitWithStatus(csl *workloadsv1alpha1.Console) error {
+	pod, _, err := c.GetAttachablePod(csl)
+	if err != nil {
+		return err
+	}
+
+	for pod.Status.Phase == "Running" {
```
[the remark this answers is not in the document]
benwh: Ah it's not so bad because we'll already have blocked on the `c.Attach` call before this (but not great that this function couldn't be called safely in isolation)

Arguably even then there's an edge case - if the SPDY stream drops for whatever reset (e.g. network blip), and the upstream 'attacher' code doesn't correctly propagate the error, then we could fall into this. Not sure how likely that is though!

pkg/workloads/console/runner/runner.go
```diff
@@ -232,11 +228,36 @@ func (c *Runner) Create(ctx context.Context, opts CreateOptions) (*workloadsv1al
...
+			return csl, err
+		}
+
+		return csl, c.exitWithStatus(csl)
 	}
 
 	return csl, nil
 }
 
+func (c *Runner) exitWithStatus(csl *workloadsv1alpha1.Console) error {
```
benwh, line 241: The naming of this thing feels odd, it doesn't do what I expect it might do from the name alone.

I'm not really sure of a better function name though - `errorIfUnsuccessfulCompletion`?

pkg/workloads/console/runner/runner.go
```diff
@@ -736,10 +769,56 @@ func (c *Runner) GetAttachablePod(csl *workloadsv1alpha1.Console) (*corev1.Pod,
...
 			return pod, c.Name, nil
 		}
 	}
 
 	return nil, "", errors.New("no attachable pod found")
 }
+
+// NoninteractiveAttacher knows how to attach to stdio of an existing container, relaying io
+// to the parent process file descriptors.
+type NoninteractiveAttacher struct {
```
benwh, line 782: Can we shift this up in the file to sit just under the `InteractiveAttacher` struct and funcs?

pkg/workloads/console/runner/runner.go
```diff
@@ -736,10 +769,56 @@ func (c *Runner) GetAttachablePod(csl *workloadsv1alpha1.Console) (*corev1.Pod,
...
+		if !csl.Spec.Interactive || c.TTY {
 			return pod, c.Name, nil
 		}
 	}
 
 	return nil, "", errors.New("no attachable pod found")
 }
+
+// NoninteractiveAttacher knows how to attach to stdio of an existing container, relaying io
+// to the parent process file descriptors.
```
benwh, line 781: This comment should probably be expanded on - it would be useful to lay out the differences between this and the interactive version

pkg/workloads/console/runner/runner.go
benwh, line 782: Did we create a new type to avoid having to break the public interface with a new parameter on the `Attach` func? Just curious really - while there is quite a bit of duplication I can definitely see an argument for either option.

pkg/workloads/console/runner/runner.go
```diff
@@ -232,11 +228,36 @@ func (c *Runner) Create(ctx context.Context, opts CreateOptions) (*workloadsv1al
 				Hook:       opts.Hook,
 			},
 		)
+		if err != nil {
+			return csl, err
+		}
+
+		return csl, c.exitWithStatus(csl)
```
benwh, line 235: Why add this here, rather than in `Runner.Attach` where it would take effect for both create and attach?

Regardless of what kubectl does I reckon it's a bit confusing to have one subcommand check for the exit status and the other to not, even more so if undocumented!

apis/workloads/v1alpha1/console_types.go
```diff
@@ -44,6 +44,12 @@ type ConsoleSpec struct {
 	// The command and arguments to execute. If not specified the command from
 	// the template specification will be used.
 	Command []string `json:"command,omitempty"`
+
+	// Enable TTY and STDIN on the underlying container, so clients can attach
+	// interactively. This should usually be set to true; however, in certain
+	// situations, enabling the TTY on a container in the console causes
+	// breakage - in Tekton steps, for example.
+	Interactive bool `json:"interactive,omitempty"`
```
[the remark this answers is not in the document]
tragiclifestories: Since this is a breaking change the idea would be to cut a new major version, and release a new version of Anu that did things correctly (and then hustle people to upgrade). 

The reason for making it a positive flag was simply that it reads better - it's weird to have lots of inverted boolean conditions around the place IMO. But it could be the other way if you'd prefer it (or, indeed, a boolean pointer).

pkg/workloads/console/runner/runner.go
```diff
@@ -736,10 +769,56 @@ func (c *Runner) GetAttachablePod(csl *workloadsv1alpha1.Console) (*corev1.Pod,
 	}
 
 	for _, c := range pod.Spec.Containers {
-		if c.TTY {
+		if !csl.Spec.Interactive || c.TTY {
```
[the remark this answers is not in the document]
tragiclifestories: Basically, if it's not an interactive console, then there's going to be no container with TTY = true by definition, so we need another heuristic. In this case, we've just done "take the first one \*shrug\*". It would probably be clearer to have a separate early return. 

I'm not sure if we want a more robust mechanism here for getting 'the right container', mind.

pkg/workloads/console/runner/runner.go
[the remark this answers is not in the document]
tragiclifestories: The reason I did this was more that `pods.get` is not guaranteed to return the most up to date view of the resource, so there may be a delay between Attach returning and the pod appearing to be finished.

I think your edge case is a serious thing to consider though, I have no idea how reliable these SPDY streamers are or if they can 'fail successfully' ... I'll see if I can reimplement this like the other wait func.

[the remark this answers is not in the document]
tragiclifestories: While adding a watcher into this logic, I discovered that the pod is never (quite) completed when we make this first call to the API to retrieve it so the watcher is very necessary!

pkg/workloads/console/runner/runner.go
tragiclifestories: I went for `waitForSuccess`

pkg/workloads/console/runner/runner.go
tragiclifestories: I think this is probably just copypasta.

pkg/workloads/console/runner/runner.go
tragiclifestories: It was a personal taste thing really - I felt that there were getting to be so many `if csl.Spec.Interactive` blocks that it was getting hard to follow the logic in any particular place.

pkg/workloads/console/runner/runner.go
tragiclifestories: I've moved it, since it doesn't make things any more complicated.

tragiclifestories:
@benwh Updated in response to comments. I was planning to write some tests but I don't think we have _any_ existing tests around attachment and I wouldn't really know how to start testing something this, er, weird.

apis/workloads/v1alpha1/console_types.go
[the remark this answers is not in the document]
benwh: > then hustle people to upgrade

Hmmm yeah, that's the painful part here - it's not going to break in a graceful way, people are still going to be able to create consoles with the old version, they'll just be broken or may violate some expectations like "I can create a detached console and then later attach to it", until they've upgraded.

Really the proper way to add this option, in its current form, would be to cut a new version of the CRDs. But we don't have a good story around that, and I wouldn't recommend trying to tackle it as part of this PR either.

> it's weird to have lots of inverted boolean conditions around the place

Yeah fair point here. I think it does depend upon how you look at this feature though - you can very easily see it as "Interactive is the default, non-interactive is an optional modifier" (but arguably that's an artefact of the project's history). We currently only check for this condition twice, and in one place we've already inverted the boolean, so it might not be that bad.

---

I'll defer the decision on this one to you and the team though - given that you'll be handling the coordination of the rollout you're probably best placed to the effort of managing users and tradeoffs of risks.

pkg/workloads/console/runner/runner.go
benwh: Just spotted that similarly, the comment above `func (a *NoninteractiveAttacher) Attach` is wrong

pkg/workloads/console/runner/runner.go
benwh: Yeah that's fair! It definitely would add a couple more conditional branches. Don't mind keeping it as is

I do wonder whether we should be exposing these structs publicly within the package or not. We're already changing the 'public interface' by removing the `Attacher` struct. Given that external code shouldn't have a reason to create a `NonInteractiveAttacher` etc maybe we should make these types private so as to allow for any future refactoring?

benwh (commented):
Apologies for the slow re-review! Couple more comments but this is getting pretty close now I think

benwh:
Agree that the tests would be a nightmare to write, so yeah let's skip those here

apis/workloads/v1alpha1/console_types.go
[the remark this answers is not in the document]
Sinjo: I'm finding it really hard to separate maybe being biased because it's how they are already from actually thinking consoles should be interactive by default.

My gut says interactive by default, but I don't have that strong a reason beyond it matching what the wrapping CLI will request by default. Not sure if that's a good reason for the library to behave that way.

pkg/workloads/console/runner/runner.go
```diff
@@ -232,30 +232,71 @@ func (c *Runner) Create(ctx context.Context, opts CreateOptions) (*workloadsv1al
...
+		return csl, c.waitForSuccess(ctx, csl)
 	}
 
 	return csl, nil
 }
 
-func (c *Runner) exitWithStatus(csl *workloadsv1alpha1.Console) error {
+func (c *Runner) waitForSuccess(ctx context.Context, csl *workloadsv1alpha1.Console) error {
+	isRunning := func(pod *corev1.Pod) bool {
+		return pod != nil && pod.Status.Phase == "Running"
```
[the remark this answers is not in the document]
tragiclifestories: Totally - I thought such things would exist but couldn't find them

pkg/workloads/console/runner/runner.go
tragiclifestories: I think that's a good idea.

pkg/workloads/console/runner/runner.go
```diff
@@ -232,30 +232,71 @@ func (c *Runner) Create(ctx context.Context, opts CreateOptions) (*workloadsv1al
...
+	if err != nil && !apierrors.IsNotFound(err) {
+		return fmt.Errorf("error retrieving console: %w", err)
+	}
+
+	if succeeded(pod) {
 		return nil
 	}
 
-	return fmt.Errorf("Pod in unsuccessful state %s: %s", pod.Status.Phase, pod.Status.Message)
+	if !isRunning(pod) {
```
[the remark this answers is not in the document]
tragiclifestories: Since this runs after Create and Attach have completed, this depends on whether a console can be in Running phase while its underlying pod is Pending - I hope not?

apis/workloads/v1alpha1/console_types.go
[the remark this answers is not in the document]
tragiclifestories: I changed the flag in CreateOptions - this is in fact the interface that Anu uses so should allow a seamless upgrade.

controllers/workloads/console/integration/integration_test.go
tragiclifestories: See https://github.com/gocardless/theatre/pull/211#discussion_r483461870

pkg/workloads/console/runner/runner.go
tragiclifestories: And worst of all, it has a missing apostrophe 😱

tragiclifestories:
@benwh Thanks a lot for reviewing - I think that's everything ...

pkg/workloads/console/runner/runner.go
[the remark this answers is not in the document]
benwh: That's fair - as long as we don't end up calling this function from anywhere else then all is good. Probably not one to worry about for the moment, as any future programming error will manifest itself in an obvious and non-dangerous way

apis/workloads/v1alpha1/console_types.go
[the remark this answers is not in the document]
benwh: Maybe I'm missing something here, but I'm not sure it does provide a seamless upgrade?

People will currently have binaries that have no knowledge of the `Interactive` field on `ConsoleSpec`, and won't be setting it in their version of the `CreateResource` function. This means that we're going to be creating `Console` objects that are missing the `Interactive` field, so setting it to implicitly false

If we do go down this route then we need to make sure that all human users are upgraded before we roll out the new controller, as well as ensuring that any automation (PS DB migrations?) still works without a TTY present (some processes get funny and perform different behaviour when a TTY isn't found)

benwh (commented):
Just noticed that there's no update to `cmd/theatre-consoles` ? We should definitely tie the new option in there (and it makes it much quicker to test changes, not having to update the dependency elsewhere!)

benwh:
We probably also want to add the field to `config/samples/workloads_v1alpha1_console.yaml` for visibility

benwh:
On the plus side, I've tested this out in the acceptance environment and all seems to work as expected!

apis/workloads/v1alpha1/console_types.go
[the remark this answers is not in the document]
tragiclifestories: ... GODDAMMIT

You're right obviously. I've pushed a change but will need to deploy the controller to lab to test.

tragiclifestories:
@benwh OK .... I think that's everything 😂 

AND THIS TIME I MEAN IT

tragiclifestories:
(Also I tested this in lab and it worked)

benwh (dismissed):
Looking good, let's do this!

I'll leave the squashing to you 😄

VERSION
```diff
@@ -1 +1 @@
-2.0.1
+2.1.0
```
tragiclifestories, line 1: @benwh I think this is the right thing? LMK if not.

anderoo (approved):

benwh (approved):
Yep version bump LGTM

PR #212, opened 2020-08-24
Remove clientset

jackatbancast:
### Remove outdated generated client
e1aa554

This changeset removes the generated client code. This is a hangover
from kubebuilder v1, and causes projects that use go.mod to fail when
importing this project.

### Update version to 2.0.1
89bc8b9

Bumping a patch as 2.0.0 doesn't currently work when importing from a
module using go.mod


theobarberbany (approved):
Seems sensible.

PR #213, opened 2020-09-07
Increment module version number

tragiclifestories:
Every time I interact with Go modules my Call of Cthulhu sanity stat takes a hit. But I _think_ this is necessary to depend on 2.x releases of theatre rather than exact commits? 

Also there's a patch version bump to trigger goreleaser.


tragiclifestories:
This is clearly going to be a horrible rabbit hole, bye now 🏃

PR #217, opened 2020-09-28
pkg/workloads/console/runner.MultipleConsoleTemplateError

lawrencejones:
Public error type that can be used to provide custom handling options
for anyone vendoring the console runner. It attaches the offending
ConsoleTemplates so the caller can access whatever metadata they want to
see, such as labels.


benwh (approved):
LGTM, very sensible

PR #220, opened 2020-11-03
envconsul: load secrets statically

jackatbancast:
This changeset adds functionality to theatre-envconsul to load secrets statically using envconsul through a helper to read the secrets environment and load it whilst skirting the leasing logic inside of envconsul.


benwh (approved):
LGTM!

PR #221, opened 2020-11-05
Add pods:watch to recommended console role rules

tragiclifestories:
This is necessary to observe the exit status of pods to which a console is attached.

Don't know if we want to motivate that - we don't seem to have motivated any of the other rules specifically.


benwh (approved):
❤️

PR #225, opened 2020-12-20
theatre-envconsul: add support for writing files

theobarberbany:
If an env var value is formatted `'vault-file:reference'` or
`'vault-file:reference:/path/to/store/file'` theatre-envconsul will now
write the secret found at the reference to either, a randomly generated
file under /tmp, or to the path provided.

### DIFF
--- a/cmd/theatre-envconsul/main.go
+++ b/cmd/theatre-envconsul/main.go
@@ -66,7 +66,7 @@ func main() {
 	defer cancel()
 
 	if err := mainError(ctx, command); err != nil {
-		logger.Error(err, "exiting with error", "error", err)
+		logger.Error(err, "exiting with error")
 		os.Exit(1)
 	}
 }
--- a/cmd/theatre-envconsul/main.go
+++ b/cmd/theatre-envconsul/main.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	execpkg "os/exec"
 	"path"
+	"path/filepath"
 	"strings"
 	"syscall"
 
@@ -133,6 +134,38 @@ func mainError(ctx context.Context, command string) (err error) {
 			}
 		}
 
+		var filePaths = environment{}
+
+		// Rewrite 'vault-file:' prefixed env vars to 'vault:' prefixed env vars. Store the
+		// paths to which they should be written to in filePaths. When no path is
+		// provided, use "" as a placeholder.
+		//
+		// For reference, the expected formats are 'vault-file:tls-key/2021010100' and
+		// 'vault-file:ssh-key/2021010100:/home/user/.ssh/id_rsa'
+		for key, value := range env {
+			if strings.HasPrefix(value, "vault-file:") {
+				trimmed := strings.TrimSpace(
+					strings.TrimPrefix(value, "vault-file:"),
+				)
+				if len(trimmed) == 0 {
+					return fmt.Errorf("empty vault-file env var: %v", value)
+				}
+
+				split := strings.SplitN(trimmed, ":", 2)
+
+				// determine if we define a path at which to place the file. For SplitN,
+				// N=2 so we only have two cases
+				switch len(split) {
+				case 2: // path and key
+					filePaths[key] = split[1]
+					env[key] = fmt.Sprintf("vault:%s", split[0])
+				case 1: // just key
+					filePaths[key] = ""
+					env[key] = fmt.Sprintf("vault:%s", trimmed)
+				}
+			}
+		}
+
 		var secretEnv = environment{}
 
 		// For all the environment values that look like they should be vault references, we
@@ -191,6 +224,40 @@ func mainError(ctx context.Context, command string) (err error) {
 		err = json.Unmarshal(output, &envMap)
 		if err != nil {
 			return errors.Wrap(err, "failed to decode envconsul environment variables")
+		}
+
+		// For every file reference in filePaths, write the value resolved by envconsul to
+		// the path in filePaths. Returns the path of the written file in the env var that
+		// requested it.
+		for key, path := range filePaths {
+			if path == "" {
+				// generate file path prefixed by key
+				tempFilePath, err := ioutil.TempFile("", fmt.Sprintf("%s-*", key))
+				if err != nil {
+					return errors.Wrap(err, fmt.Sprintf("failed to write temporary file for key %s", key))
+				}
+
+				path = tempFilePath.Name()
+			}
+			// ensure the path structure is available
+			err := os.MkdirAll(filepath.Dir(path), 0600)
+			if err != nil {
+				return fmt.Errorf("failed to ensure path structure is available: %s", err.Error())
+			}
+
+			logger.Info(
+				"creating vault secret file",
+				"event", "envconsul_secret_file.create",
+				"path", path,
+			)
+			// write file with value of envMap[key]
+			if err := ioutil.WriteFile(path, []byte(envMap[key]), 0600); err != nil {
+				return errors.Wrap(err,
+					fmt.Sprintf("failed to write file with key %s to path %s", key, path))
+			}
+
+			// update the env with the location of the file we've written
+			envMap[key] = path
 		}
 
 		// Update the environment variables based on updated environment variables
--- a/cmd/theatre-envconsul/main.go
+++ b/cmd/theatre-envconsul/main.go
@@ -217,7 +217,11 @@ func mainError(ctx context.Context, command string) (err error) {
 
 		output, err := execpkg.CommandContext(ctx, envconsulBinaryPath, envconsulArgs...).Output()
 		if err != nil {
-			return errors.Wrap(err, "failed to get envconsul environment variables")
+			if ee, ok := err.(*execpkg.ExitError); ok {
+				output = ee.Stderr
+			}
+
+			return errors.Wrapf(err, "failed to get envconsul environment variables: %s", output)
 		}
 
 		envMap := map[string]string{}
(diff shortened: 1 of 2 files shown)


cmd/theatre-envconsul/main.go
```diff
@@ -193,6 +219,25 @@ func mainError(ctx context.Context, command string) (err error) {
...
+				// generate file path prefixed by key
+				tempFilePath, err := ioutil.TempFile("", fmt.Sprintf("%s-*", key))
+				if err != nil {
+					return errors.Wrap(err, fmt.Sprintf("failed to write temporary file for key %s", key))
+				}
+				value = tempFilePath.Name()
+			}
+
+			// write file with value of envMap[key]
+			if err := ioutil.WriteFile(value, []byte(env[key]), 0644); err != nil {
```
theobarberbany, line 235: this does not work if value (path) is nested e.g /tmp/secrets/secretfile.

cmd/theatre-envconsul/main.go
```diff
@@ -193,6 +219,25 @@ func mainError(ctx context.Context, command string) (err error) {
 			return errors.Wrap(err, "failed to decode envconsul environment variables")
 		}
 
+		// For every file reference, write the value resolved by envconsul to the path in
+		// filePaths
+		for key, value := range filePaths {
```
theobarberbany, line 224: /s/value/path/g

cmd/theatre-envconsul/main.go
```diff
@@ -133,6 +133,32 @@ func mainError(ctx context.Context, command string) (err error) {
 			}
 		}
 
+		var filePaths = environment{}
+		var pathTBD = "PATH_TBD"
```
benwh, line 137: Does this make more sense as a const? Or can we use a more idiomatic sentinel value such as empty string?

cmd/theatre-envconsul/main.go
```diff
@@ -133,6 +133,32 @@ func mainError(ctx context.Context, command string) (err error) {
...
+				trimmed := strings.TrimPrefix(value, "vault-file:")
+				split := strings.SplitN(trimmed, ":", 2)
+				switch len(split) {
+				case 2: // path and key
+					filePaths[key] = split[1]
+					env[key] = fmt.Sprintf("vault:%s", split[0])
+				case 1: // just key
+					filePaths[key] = pathTBD
+					env[key] = fmt.Sprintf("vault:%s", trimmed)
+				default:
```
benwh, line 153: Given what's returned from `SplitN`, I don't think that we can ever actually reach this?

cmd/theatre-envconsul/main.go
benwh: WDYT about using `os.MkdirAll` before this, to ensure that the path structure is available?

benwh, line 235: Shouldn't this be `envMap[key]`?

cmd/theatre-envconsul/main.go
```diff
@@ -193,6 +219,25 @@ func mainError(ctx context.Context, command string) (err error) {
...
+			if value == pathTBD {
+				// generate file path prefixed by key
+				tempFilePath, err := ioutil.TempFile("", fmt.Sprintf("%s-*", key))
+				if err != nil {
+					return errors.Wrap(err, fmt.Sprintf("failed to write temporary file for key %s", key))
+				}
+				value = tempFilePath.Name()
+			}
+
+			// write file with value of envMap[key]
```
benwh, line 234: Once we've done this, should we *unset* the environment variables that correspond to the `vault-file:KEY:/path` values, so that they're not exposed to the app that's exec'd?

cmd/theatre-envconsul/main.go
theobarberbany: I think an empty string would work!

cmd/theatre-envconsul/main.go
theobarberbany: I think we could do that, or would it be an idea to set the value of the env var with the path where the secret now is written?

benwh: I'd sort of assumed that we'd only be doing that for ones where we randomly generate the path. But actually, maybe this does make sense to do always!

cmd/theatre-envconsul/main.go
theobarberbany: Ah yea, no `:` present will evaluate to `case 1`, which is what we want. I suppose with `SplitN(trimmed, ":", -1)` we have validation for too many `':'`'s? Or do you think it should just be dropped?

benwh: I think arguably we could just drop it. We may want some separate validation for when the input string is simply `vault-file:` with no extra components though. If that's set, then we should probably just throw an error and exit

cmd/theatre-envconsul/main.go
```diff
@@ -193,6 +220,30 @@ func mainError(ctx context.Context, command string) (err error) {
...
+				if err != nil {
+					return errors.Wrap(err, fmt.Sprintf("failed to write temporary file for key %s", key))
+				}
+				path = tempFilePath.Name()
+			}
+			// ensure the path structure is available
+			os.MkdirAll(filepath.Dir(path), 0644)
+
+			// write file with value of envMap[key]
+			if err := ioutil.WriteFile(path, []byte(envMap[key]), 0644); err != nil {
```
benwh, line 238: For observability and debugging it might be nice to write out a log line whenever we write a file to disk

theobarberbany: I've copied the schema of the previous log messages.

cmd/theatre-envconsul/main.go
```diff
@@ -133,6 +134,37 @@ func mainError(ctx context.Context, command string) (err error) {
...
+				split := strings.SplitN(trimmed, ":", 2)
+				switch len(split) {
+				case 2: // path and key
+					filePaths[key] = split[1]
+					env[key] = fmt.Sprintf("vault:%s", split[0])
+				case 1: // just key
+					filePaths[key] = pathTBD
+					env[key] = fmt.Sprintf("vault:%s", trimmed)
+				default:
+					// will not execute because of SplitN, N=2
```
benwh, line 163: I think it's less confusing to just remove this `default` case, and move the comment above the `switch` and amend appropriately

cmd/theatre-envconsul/main.go
```diff
@@ -193,6 +225,35 @@ func mainError(ctx context.Context, command string) (err error) {
...
+		// For every file reference, write the value resolved by envconsul to the path in
+		// filePaths
+		for key, path := range filePaths {
+			if path == pathTBD {
+				// generate file path prefixed by key
+				tempFilePath, err := ioutil.TempFile("", fmt.Sprintf("%s-*", key))
+				if err != nil {
+					return errors.Wrap(err, fmt.Sprintf("failed to write temporary file for key %s", key))
+				}
+				path = tempFilePath.Name()
```
benwh, line 237: Add newline above for readability

cmd/theatre-envconsul/main.go
```diff
@@ -193,6 +225,35 @@ func mainError(ctx context.Context, command string) (err error) {
...
+			}
+			// ensure the path structure is available
+			os.MkdirAll(filepath.Dir(path), 0644)
```
benwh, line 240: We're not checking the returned error from this call. It can blow up, e.g. in the case that the path exists but is a file rather than a dir.

cmd/theatre-envconsul/main.go
```diff
@@ -133,6 +134,37 @@ func mainError(ctx context.Context, command string) (err error) {
...
+		for key, value := range env {
+			if strings.HasPrefix(value, "vault-file:") {
+				// determine if we define a path at which to place the file
+				trimmed := strings.TrimSpace(
+					strings.TrimPrefix(value, "vault-file:"),
+				)
+				if len(trimmed) == 0 {
+					return errors.Wrap(
+						fmt.Errorf("empty vault-file env var: %v", value),
+						"failed parse vault-file env var")
```
benwh, line 151: The wrapping is confusing and doesn't really give us anything, so I'd suggest:

```suggestion
					return fmt.Errorf("empty vault-file env var: %v", value)
```

cmd/theatre-envconsul/main.go
```diff
@@ -133,6 +134,37 @@ func mainError(ctx context.Context, command string) (err error) {
 			}
 		}
 
+		var filePaths = environment{}
+		const pathTBD = ""
+
+		// Rewrite vault-file: envs to vault: envs. Store the paths to which they should
```
benwh, line 140: Would be good in this comment to add some examples of what it actually looks like, for clarity

e.g.
`vault-file:tls-key/2021010100`
`vault-file:ssh-key/2021010100:/home/user/.ssh/id_rsa`

benwh (commented):
Couple of small fixes to make, most importantly the file permissions, but otherwise this looks good. Have tested it out locally and works as expected!

Pushing a couple of small commits that I found helped when testing/debugging this.

benwh:
We should also exercise this within the vault acceptance tests, which is confusingly located in a different part of the hierarchy: https://github.com/gocardless/theatre/blob/master/cmd/vault-manager/acceptance/acceptance.go

cmd/theatre-envconsul/main.go
```diff
@@ -133,6 +134,32 @@ func mainError(ctx context.Context, command string) (err error) {
 			}
 		}
 
+		var filePaths = environment{}
+		const pathTBD = ""
```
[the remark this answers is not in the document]
theobarberbany: Oh yea, I was thinking I'd leave it in for understandability, but with a comment this is just as good

benwh (approved):
LGTM

PR #226, opened 2021-02-24
Remove envconsul dependency from theatre-envconsul

benwh:
Previously we've shelled-out to the `envconsul` binary in order to
retrieve our secret material from Vault. Through this we've discovered
several shortcomings:
- Failure to provision 'secret files' which have large/non-ASCII bodies.
- Failure to provision environment variables which are referencing the
  same Vault key path.
- Hanging when a process is wrapped with `envconsul` but there's no
  secret material to fetch.
- Workarounds required for shellwords splitting.

By removing `envconsul` and implementing the fetching from Vault
ourselves we simplify the code significantly and end up with something
that's more performant and easy to reason about, as well as fixing the
above issues.


benwh (commented):
The follow up commits look good to me!

theobarberbany (approved):

PR #231, opened 2021-03-16
Fix detaching from interactive consoles

dyson:
Currently detaching from a console is not behaving as expected and we
appear to hang once detached. What happens is the following:

    You create an interactive console with a tty
    You detach using CTRL-p CTRL-q
    The cli is now stuck in c.waitForSuccess as the console is in a running state
    You can CTRL-c to kill the cli to escape

To resolve this we shouldn't be waiting for success after we exit from
an interactive console (either from completion or detaching). This used
to be the case but the behaviour changed when support for attaching to
non-interactive consoles (this streams the consoles stdout) was added.


benwh (approved):
LGTM

PR #233, opened 2021-03-17
V3.0.2

dyson:
We bumped to v3 without bumping the go.mod and codebase which prevented
importing theatre as a dependency into other codebases. This resolves that.

go.mod has already been updated and merged to master.


benwh (approved):
LGTM

PR #234, opened 2021-03-17
Allow config of theatre-secrets debug and timeout

benwh:
7cdd17b: Allow config of theatre-secrets debug and timeout

Provide new flags for the vault-manager binary that allow us to decide
whether the `--debug` flag is passed and the `--vault-http-timeout` flag
can be set, when invoking the `theatre-secrets` wrapper.

This allows us to temporarily enable debug logging in order to determine
the durations of requests to Vault, as well as bump up the timeout if
required.

3ffe458: Add debug logging to theatre-secrets

This allows us to flip debug logging on and see how long our requests
are taking, and whether they're close to timing out.

Debug logging is a bit odd here. The `logr` API doesn't actually include
a `Debug()` function, you need to use [V-levels][0], and the interaction with
Zap is described [here][1].

[0]: https://github.com/go-logr/logr#how-do-i-choose-my-v-levels
[1]: https://github.com/go-logr/zapr/blob/master/zapr.go#L50-L53


benwh:
This has been deployed to `compute-lab` with the debug flag enabled on the statefulset.

dyson (approved):

PR #237, opened 2021-08-02
Replace slash and period with underscore for label names

jackatbancast:
This change replaces all occurences of `.` and `/` in label
keys. Occasionally a kubernetes label will take the form of a url,
e.g. `app.kubernetes.io/instance`. In this case we would output that
key prefixed by our given parameter.

This can cause issues with systems such as ElasticSearch where it will
see the `.` as an object path separator and attempt to build the
object from the labels. This may work but is likely unintended
behaviour and can be surprising.

This issue is compounded if you have a matching key to the first
segment of the domain-prefixed key, e.g. `app` in our case. This would
conflict with the object that is being created by `app.kubernetes.io`
and potentially cause errors to occur.

By reducing the label key to a form similar to snake case we can
avoid this issue and provide dependant systems with a label form that
doesn't conflict. Additionaly we can provide more consistent behaviour.

### DIFF
--- a/pkg/logging/labels.go
+++ b/pkg/logging/labels.go
@@ -1,6 +1,9 @@
 package logging
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/go-logr/logr"
 )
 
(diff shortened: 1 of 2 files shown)


VSpike (dismissed):

pkg/logging/labels.go
```diff
@@ -1,15 +1,44 @@
...
+//
+// This behaviour matches the behaviour seen in some other systems
+// such as Prometheus where labels are reduced to a alphanumeric
+// string with underscore spacers.
+//
+// e.g. app.kubernetes.io/instance would become
+// app_kubernetes_io_instance
+func linearLabelKey(labelKey string) string {
+	labelKey = strings.ReplaceAll(labelKey, "/", "_")
+	labelKey = strings.ReplaceAll(labelKey, ".", "_")
```
benwh, line 41: I wondered if we should be doing the same for dashes

https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

But actually it [seems like](https://discuss.elastic.co/t/legal-character-set-for-field-names/190796) ES would be fine with these, so no issue there

benwh (approved):
LGTM

pkg/logging/labels.go
```diff
@@ -1,15 +1,44 @@
...
 )
 
 // WithLabels decorates a logr.Logger so that any log entries contain all
 // labels which keys are prefix by labelKeyPrefix
 func WithLabels(logger logr.Logger, labels map[string]string, labelKeyPrefix string) logr.Logger {
 	for key, value := range labels {
-		logger = logger.WithValues(labelKeyPrefix+key, value)
+		logger = logger.WithValues(
+			fmt.Sprintf(
+				"%s_%s",
```
jackatbancast, line 16: Had a quick check and I think we might want to remove the underscore here, we already add this as part of the prefix.

```suggestion
				"%s%s",
```

theobarberbany (approved):

PR #242, opened 2021-11-24
Add console observer webhook

jackatbancast:
This changeset adds a webhook that will attempt to observe console
attach events. This is achieved by watching the pod/attach events via
a validating webhook.

By Ignoring the failure case of this webhook we can ensure that we are
optimistically observing attach events to pods, whilst removing the
possibility of blocking console creation. This is useful when there is
an incident to ensure engineers continue to have access.


benwh (approved):
LGTM

PR #243, opened 2021-11-24
CI-999: Modify console template on creation to add tlog

VSpike:
This PR addresses CI-999 (adding tlog into the console job template on creation) and CI-1010 (adding a sidecar to the console to read the session data and stream to Pubsub).

It modifies the console operator to do the following:

* Add a volume mount for the log files
* Add the volume mount to the main container(s)
* Modify the main container(s) commands to wrap everything in tlog
* Add a sidecar with the appropriate image
* Pass through configuration data as env vars to the sidecar
* Create a role and rolebinding that will allow the console pod to read information about itself


controllers/workloads/console/controller.go
```diff
@@ -592,6 +653,126 @@ func requeueAfterInterval(logger logr.Logger, interval time.Duration) reconcile.
...
+	}
+	if len(r.SessionSidecarImage) > 0 {
+		mutatedTemplate.Spec.Containers = append(
+			mutatedTemplate.Spec.Containers,
+			r.buildSidecarContainer(),
+		)
+		// The grace period for the pod must be longer than Sidewrap's delay
+		// and grace periods combined
+		var gracePeriod int64 = SidewrapShutdownDelay + SidewrapGracePeriod + 30
+		mutatedTemplate.Spec.TerminationGracePeriodSeconds = &gracePeriod
```
benwh, line 764: Is there a risk that we could be decreasing the grace period here - should we guard against it?

controllers/workloads/console/controller.go
```diff
@@ -117,6 +135,57 @@ func (r *ConsoleReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manag
...
+	}
+
+	// Create or update the service role
+	role := buildServiceRole(rbacName, csl.Status.PodName)
+	if err := r.createOrUpdate(ctx, logger, csl, role, Role, recutil.RoleDiff); err != nil {
+		return err
+	}
+
+	// Create or update the service rolebinding
+	drb := buildServiceRoleBinding(rbacName, role, tpl.Spec.Template.Spec.ServiceAccountName)
```
benwh, line 181: Is `ServiceAccountName` guaranteed to be non-empty?

controllers/workloads/console/controller.go
```diff
@@ -676,7 +862,46 @@ func (r *ConsoleReconciler) buildJob(logger logr.Logger, name types.NamespacedNa
 	}
 }
 
-func buildRole(name types.NamespacedName, podName string) *rbacv1.Role {
+func buildServiceRole(name types.NamespacedName, podName string) *rbacv1.Role {
```
benwh, line 865: Some comments about why we need to build and apply these roles/rolebindings wouldn't go amiss

benwh: Maybe the comment could go on `createOrUpdateServiceRbac` though

controllers/workloads/console/controller.go
```diff
@@ -592,6 +653,126 @@ func requeueAfterInterval(logger logr.Logger, interval time.Duration) reconcile.
...
+				ReadOnly:  false,
+			},
+		},
+		Resources: corev1.ResourceRequirements{
+			Limits: corev1.ResourceList{
+				corev1.ResourceMemory: resource.MustParse("512Mi"),
+				corev1.ResourceCPU:    resource.MustParse("512m"),
+			},
+			Requests: corev1.ResourceList{
+				corev1.ResourceMemory: resource.MustParse("512Mi"),
```
benwh, line 678: These seem pretty beefy, but happy to go with it and tune it down later on if we see that it's excessive

controllers/workloads/console/controller.go
```diff
@@ -660,14 +841,19 @@ func (r *ConsoleReconciler) buildJob(logger logr.Logger, name types.NamespacedNa
 		jobTemplate.ObjectMeta.Labels,
 	)
 
+	podTemplate := (*corev1.PodTemplateSpec)(jobTemplate)
+	if r.EnableSessionRecording {
+		podTemplate = r.modifyPodTemplate(logger, podTemplate)
```
benwh, line 846: `modifyPodTemplate` is arguably a bit of a generic name, given that it's only relevant within the context of session recording

controllers/workloads/console/controller.go
```diff
@@ -592,6 +653,126 @@ func requeueAfterInterval(logger logr.Logger, interval time.Duration) reconcile.
...
+	// session recording wrapper which spawns the original
+	// command
+	for ix := range mutatedTemplate.Spec.Containers {
+		ctr := &mutatedTemplate.Spec.Containers[ix]
+		execCommand := []string{"--"}
+		execCommand = append(execCommand, ctr.Command...)
+		execCommand = append(execCommand, ctr.Args...)
+
+		ctr.Command = []string{"tlog-rec"}
+		args := []string{"-o", sessionRecordFileName(ix)}
```
benwh, line 744: I think we may also want to specify `--log-input=true`, so that we capture stdin separately (the default is false)

controllers/workloads/console/controller.go
```diff
@@ -592,6 +653,126 @@ func requeueAfterInterval(logger logr.Logger, interval time.Duration) reconcile.
...
+		mutatedTemplate.Spec.Containers = append(
+			mutatedTemplate.Spec.Containers,
+			r.buildSidecarContainer(),
+		)
+		// The grace period for the pod must be longer than Sidewrap's delay
+		// and grace periods combined
+		var gracePeriod int64 = SidewrapShutdownDelay + SidewrapGracePeriod + 30
+		mutatedTemplate.Spec.TerminationGracePeriodSeconds = &gracePeriod
+	} else {
+		msg := "Unable to add session recording sidecar due to missing image setting"
```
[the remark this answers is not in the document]
benwh: Ah, now I see that it's maybe intended to be optional. What's the use-case for that?

benwh (commented):
Looks good overall - some minor comments

controllers/workloads/console/controller.go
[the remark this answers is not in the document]
VSpike: Yeah, I think you're right. I'll add some verification at the top of `main.go`

controllers/workloads/console/controller.go
VSpike: Good catch, I'll add a check for that case.

controllers/workloads/console/controller.go
VSpike: I think so!! Do you know if there's a case where it might be empty? Wouldn't a service always have an associated service account that the console would run under?

controllers/workloads/console/controller.go
VSpike: Great catch, thanks

controllers/workloads/console/controller.go
VSpike: 👍

controllers/workloads/console/controller.go
VSpike: These are definitely generous but we're not sure what the actual requirements are yet.

controllers/workloads/console/controller.go
VSpike: I have now handled the case of a missing service account name!

benwh (dismissed):
LGTM

benwh (dismissed):

benwh (approved):
LGTM after squash

PR #244, opened 2021-11-24
CI-1029: Add attach support to theatre-runner authorise

VSpike:
Consoles can be attached when authorised (if allowed)

### DIFF
--- a/controllers/workloads/console/controller.go
+++ b/controllers/workloads/console/controller.go
@@ -244,6 +244,10 @@ func (r *ConsoleReconciler) Reconcile(logger logr.Logger, ctx context.Context, r
 			tpl.Spec.AdditionalAttachSubjects,
 			rbacv1.Subject{Kind: "User", Name: csl.Spec.User},
 		)
+		// Append all the authorising users to allow them to attach
+		if authorisation != nil {
+			subjects = append(subjects, authorisation.Spec.Authorisations...)
+		}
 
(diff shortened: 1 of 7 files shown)


controllers/workloads/console/controller.go
```diff
@@ -244,6 +244,8 @@ func (r *ConsoleReconciler) Reconcile(logger logr.Logger, ctx context.Context, r
 			tpl.Spec.AdditionalAttachSubjects,
 			rbacv1.Subject{Kind: "User", Name: csl.Spec.User},
 		)
+		// Append all the authorising users to allow them to attach
+		subjects = append(subjects, authorisation.Spec.Authorisations...)
```
benwh, line 248: This is fine I think, but should be considered a breaking change, as it changes the default behaviour.

We should probably update this with the details, in a follow-up: https://github.com/gocardless/theatre/blob/0e9a7f2e2c86a4701d726a0a8a37a243515f3030/controllers/workloads/console/README.md#L22

And if we end up writing any releases notes when cutting the new release, then call it out in there.

benwh: It does also seem to be causing a nil pointer dereference in the tests, so that needs to be looked at.

benwh (commented):
Looking good overall, but just needs the test failure fixup

controllers/workloads/console/controller.go
VSpike: Nil pointer deref should now be cured

benwh (approved):
LGTM

PR #247, opened 2021-11-29
Add console lifecycle event recording

jackatbancast:
This changeset adds the ability to publish events from a consoles'
lifecycle to a Google Pub/Sub topic.

This is disabled by default, and can be enabled selectively by
providing both the id for the project containing the topic, alongside
the topic id.

---

This change additionally bumps the version of the google api client
used. This requires a corresponding update in the base path of the
requests to use subdomains specific to each google api set.
e.g. www.googleapis.com to admin.googleapis.com


apis/workloads/v1alpha1/lifecycle_recorder.go
```diff
@@ -0,0 +1,147 @@
...
+
+	"github.com/go-logr/logr"
+	"github.com/gocardless/theatre/v3/pkg/workloads/console/events"
+)
+
+func CommonEventFromConsole(ctx string, eventKind events.EventKind, csl *Console) events.CommonEvent {
+	return events.CommonEvent{
+		Version:    "v1alpha1",
+		Kind:       events.KindConsole,
+		Event:      events.EventRequest,
```
benwh, line 15: Should this be `eventKind`?

pkg/workloads/console/events/publisher.go
```diff
@@ -0,0 +1,24 @@
...
+
+type Publisher interface {
+	Healther
+	Publish(context.Context, interface{}) (string, error)
+}
+
+var _ Publisher = &NopPublisher{}
+
+// https://en.wikipedia.org/wiki/NOP_(code)
+type NopPublisher struct{}
```
benwh, line 17: Interesting - I'd usually tend towards `Noop` in this kind of context, but this also feels right

pkg/workloads/console/events/events.go
```diff
@@ -5,27 +5,46 @@ import (
...
+type Kind string
+
+const (
+	KindConsole Kind = "Console"
+)
+
+type EventKind string
+
+const (
+	EventRequest    EventKind = "Requested"
```
benwh, line 17: `Request` - to match the others? (Or `Authorised`, etc.)

pkg/workloads/console/events/gcp_pubsub.go
```diff
@@ -0,0 +1,69 @@
...
+	client *pubsub.Client
+	topic  *pubsub.Topic
+}
+
+var _ Publisher = &GooglePubSubPublisher{}
+
+func NewGooglePubSubPublisher(ctx context.Context, projectName string, topicName string) (*GooglePubSubPublisher, error) {
+	client, err := pubsub.NewClient(ctx, projectName)
+	if err != nil {
+		return nil, err
```
benwh, line 31: Not sure what kind of errors we'd get back from the library, so some error wrapping might not go amiss here (and below in the same file) to keep things understandable when it goes wrong.

pkg/workloads/console/events/gcp_pubsub.go
```diff
@@ -0,0 +1,69 @@
...
+
+	result := p.topic.Publish(ctx, &pubsub.Message{Data: messageBytes})
+	id, err := result.Get(ctx)
+	if err != nil {
+		return "", err
+	}
+	return id, nil
+}
+
+func (p *GooglePubSubPublisher) Health(ctx context.Context) error {
```
benwh, line 58: What uses this?

apis/workloads/v1alpha1/console_authenticator_webhook.go
```diff
@@ -51,5 +54,13 @@ func (c *ConsoleAuthenticatorWebhook) Handle(ctx context.Context, req admission.
 
 	logger.Info(fmt.Sprintf("authentication successful for user %s", user), "event", "authentication.success", "user", user)
 
+	// This endpoint is only called when we are creating console
+	// objects. If we are succeeding the request we are creating a
+	// console request.
+	err = c.lifecycleRecorder.ConsoleRequest(ctx, csl)
+	if err != nil {
+		logging.WithNoRecord(logger).Error(err, "failed to record event", "event", "console.request")
```
benwh, line 62: As an addition later on, it would be cool to increment a Prometheus metric when we hit this condition, so that we can alert on it.

benwh (commented):
One bug, some minor comments, but overall looks decent

apis/workloads/v1alpha1/lifecycle_recorder.go
jackatbancast: Thanks for catching this. Updating this.

pkg/workloads/console/events/publisher.go
jackatbancast: I did consider a few names here but opted to more closely match the instruction name (for no particular reason).

pkg/workloads/console/events/gcp_pubsub.go
jackatbancast: Nothing uses this yet, and I'm happy to omit this from this changeset to reduce the scope of the changes.
This was more around trying to build a nice health reporting standard to our theatre operators, and using this changeset as a place to experiment.

pkg/workloads/console/events/publisher.go
dyson: TIL

pkg/workloads/console/events/gcp_pubsub.go
dyson: Yeah I think lets omit from this.

dyson (changes requested):
Just those few things Ben already raised as well as omitting the unused health code from here and I think this is looking good.

apis/workloads/v1alpha1/console_authenticator_webhook.go
jackatbancast: We should. I think this would be a nice discussion to have more generally throught the theatre codebase as well.

benwh (dismissed):
LGTM

dyson (approved):

PR #251, opened 2021-12-05
Include number of required authorisations in Pub/Sub console creation message

ttamimi:
This commit updates to the pub/sub message schema for console creation to include
the number of required authorisations. The format of the message ID is also changed
to use slashes as separators between the segments of the ID string.


apis/workloads/v1alpha1/helpers.go
```diff
@@ -214,11 +214,8 @@ matchRule:
 // HasAuthorisationRules defines whether a console template has authorisation
 // rules defined on it.
 func (ct *ConsoleTemplate) HasAuthorisationRules() bool {
-	if len(ct.Spec.AuthorisationRules) > 0 || ct.Spec.DefaultAuthorisationRule != nil {
-		return true
-	}
-
-	return false
+	return len(ct.Spec.AuthorisationRules) > 0 ||
+		(ct.Spec.DefaultAuthorisationRule != nil && ct.Spec.DefaultAuthorisationRule.AuthorisationsRequired > 0)
```
jackatbancast, line 218: I'm not convinced that we should be changing the behaviour of this function. It's also not related to the change in behaviour we are intending with this changeset.

In practice I think we should be moving to a place where there is always a default authorisation rule, even if this means defaulting the values to `0` to maintain backwards compatibility. At that point I think we should be removing this function in favour of more consistent logic to get the number of authorisations required for any given console.

apis/workloads/v1alpha1/lifecycle_recorder.go
```diff
@@ -52,18 +52,24 @@ func NewLifecycleEventRecorder(contextName string, logger logr.Logger, publisher
 	}
 }
 
-func (l *lifecycleEventRecorderImpl) ConsoleRequest(ctx context.Context, csl *Console) error {
+func (l *lifecycleEventRecorderImpl) ConsoleRequest(ctx context.Context, csl *Console, authRule *ConsoleAuthorisationRule) error {
+	var authCount int
```
jackatbancast, line 56:
```suggestion
	authCount := 0
```

I think it might be better to be explicit about the default value we're expecting here.

controllers/workloads/console/controller.go
```diff
@@ -192,6 +183,15 @@ func (r *ConsoleReconciler) Reconcile(logger logr.Logger, ctx context.Context, r
...
 
+	// If we have yet to set the owner reference record this as a
+	// new console request
+	if len(csl.OwnerReferences) == 0 {
+		err := r.LifecycleRecorder.ConsoleRequest(ctx, csl, authRule)
+		if err != nil {
+			logging.WithNoRecord(logger).Error(err, "failed to record event", "event", "console.request")
+		}
+	}
+
```
jackatbancast, line 194: Will this ever fire?
Just after where this has moved from we set the console owner, getting a new `csl` object.

```suggestion
	// If we have yet to set the owner reference record this as a
	// new console request
	if !hasOwnerReferences {
		err := r.LifecycleRecorder.ConsoleRequest(ctx, csl, authRule)
		if err != nil {
			logging.WithNoRecord(logger).Error(err, "failed to record event", "event", "console.request")
		}
	}

```

We can then capture this variable in the location this section has moved from.

```go
hasOwnerReferences := len(csl.OwnerReferences) != 0
```

This should be consistent unless we get an error on the first console request after setting the owner references, but before we reach this event emission point.
We should be aware that this, whilst unlikely, is possible if the kubernetes API should become temporarily unavialable.

ttamimi: Oh yes, very well spotted!! This is definitely a bug!

jackatbancast (approved):
This is looking really good now. Nice work!

PR #253, opened 2021-12-13
CI-1059: Add extra sidecar variables

VSpike:
Add environment variables with extra data for the session
recording sidecar for log message enrichment, including:

* Console ID
* Environment
* Service
* User
* Console Name


apis/workloads/v1alpha1/lifecycle_recorder.go
```diff
@@ -24,6 +24,7 @@ func CommonEventFromConsole(ctx string, eventKind events.EventKind, csl *Console
 
 // +kubebuilder:object:generate=false
 type LifecycleEventRecorder interface {
+	GetIdForConsole(string, string, time.Time) string
```
jackatbancast, line 27: I don't feel like this quite belongs inside of this interface. I **do** think we should create a top-level function that does this in the package, wrapping the one we implement in the `pkg/workloads/console/events` package with helpers accepting a console object.

I'm happy to be persuaded but my thoughts are that we should keep the event recorder interface fairly lean and as agnostic to the implementation details as possible, whilst adding these helpers to the implementation itself.
I don't believe the event recorder interface should have much of an idea about what an event ID entails.

benwh (approved):
LGTM

PR #254, opened 2021-12-14
CI-1059: small refactor to code to set min grace period

VSpike:
It's a minor thing, but I think I slightly prefer this version


benwh (approved):
Why not!

PR #255, opened 2021-12-16
expose metrics for sidecar container

theobarberbany:
This will expose metrics for the sidecar container. We use ports 8080
and 8090 as these are the metrics ports for PubSubTle and SideWrap
respectively.

This should allow us to scrape metrics about the health of the
two containers that power console session recording.


VSpike (approved):

benwh (commented):
Generally the convention is to start the ports at 9100 and above

https://github.com/prometheus/prometheus/wiki/Default-port-allocations

No big deal here as it's a containerised environment, but port 8080 does slightly violate the "principle of least surprise" 😄

PR #256, opened 2021-12-20
Fix metrics port names

theobarberbany:
Shorten the port names for gathering metrics from the wrapped process
and sidecar. This is because kubernetes imposes a 15 character limit,
and we are currently over.


benwh (approved):
LGTM

Didn't know about this limitation, that's an interesting one

VSpike (approved):

theobarberbany:
@benwh yeah, neither did I. 

I'm going to guess it's because every port gets a DNS record `_my-port-name._my-port-protocol.my-svc.my-namespace.svc.cluster-domain.example`

PR #261, opened 2022-02-24
Fix the getting started instructions in the README

VSpike:
The README has the wrong download URL for Kubebuilder. Update it with
the correct one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,8 +78,8 @@ all the necessary dependencies:
 ```shell
 brew cask install docker
 brew install go kubernetes-cli kustomize kind
-mkdir /usr/local/kubebuilder
-curl -fsL https://go.kubebuilder.io/dl/2.3.1/$(go env GOOS)/$(go env GOARCH) | tar -xvz --strip=1 -C /usr/local/kubebuilder
+sudo mkdir /usr/local/kubebuilder
+curl -fsL "https://github.com/kubernetes-sigs/kubebuilder/releases/download/v2.3.1/kubebuilder_2.3.1_$(go env GOOS)_$(go env GOARCH).tar.gz" | sudo tar -xvz --strip=1 -C /usr/local/kubebuilder
 export KUBEBUILDER_ASSETS=/usr/local/kubebuilder/bin
 ```
 


ttamimi (approved):

README.md
```diff
@@ -79,7 +79,7 @@ all the necessary dependencies:
 brew cask install docker
 brew install go kubernetes-cli kustomize kind
 mkdir /usr/local/kubebuilder
-curl -fsL https://go.kubebuilder.io/dl/2.3.1/$(go env GOOS)/$(go env GOARCH) | tar -xvz --strip=1 -C /usr/local/kubebuilder
+curl -fsL https://github.com/kubernetes-sigs/kubebuilder/releases/download/v2.3.1/kubebuilder_2.3.1_$(go env GOOS)_$(go env GOARCH).tar.gz | tar -xvz --strip=1 -C /usr/local/kubebuilder
```
ttamimi, line 82: Might be helpful to make it `sudo tar ...` ?

README.md
```diff
@@ -79,7 +79,7 @@ all the necessary dependencies:
 brew cask install docker
 brew install go kubernetes-cli kustomize kind
 mkdir /usr/local/kubebuilder
```
ttamimi, line 81: We should add `sudo` here, no?

ttamimi (changes requested):

README.md
EsmeAR, line 82: Annoyingly just ran this and it doesn't work in zsh I get an error saying ``parse error near `)' ``I got it to work using the fix here https://stackoverflow.com/questions/25614613/how-to-disable-zsh-substitution-autocomplete-with-url-and-backslashes which is just adding some config to oh-my-zsh (no everyone will use this though). Maybe worth adding something about this here.

EsmeAR: that does the trick! maybe just change it to be quoted and then bash and zsh are both happy 😇

VSpike: Quoting it is probably a good idea anyway!

ttamimi (approved):

PR #262, opened 2022-02-27
Emit console termination events

ttamimi:
With this change, the workloads controller will publish ConsoleTerminatedEvent
messages to Pub/Sub. The event will include per-container messages capturing
things like exit code, OS signal received, termination reason, and output
message (if any of these are available).


apis/workloads/v1alpha1/lifecycle_recorder.go
```diff
@@ -208,3 +221,38 @@ func (l *lifecycleEventRecorderImpl) ConsoleTerminate(ctx context.Context, csl *
 	l.logger.Info("event recorded", "id", id, "event", events.EventTerminated)
 	return nil
 }
+
+func appendStatusMessages(result map[string]string, containers []corev1.ContainerStatus) {
```
theobarberbany, line 225:
```suggestion
func appendStatusMessages(result map[string]string, containerStatuses []corev1.ContainerStatus) {
```

Nit: Maybe this fits more naturally?

apis/workloads/v1alpha1/lifecycle_recorder.go
```diff
@@ -208,3 +221,38 @@ func (l *lifecycleEventRecorderImpl) ConsoleTerminate(ctx context.Context, csl *
...
+	if containers == nil {
+		return
+	}
+
+	for _, containerStatus := range containers {
+		if containerStatus.State.Terminated != nil {
+			s := containerStatus.State.Terminated
```
theobarberbany, line 232: What is the state of a container that is in crashback loopoff? Will it be caught by this? Or is it handled by the waiting state below?

theobarberbany: I've just checked: looks like it'll be caught by waiting below

theobarberbany (approved):

theobarberbany:
I think this looks good! Let's test it out :)

apis/workloads/v1alpha1/lifecycle_recorder.go
ttamimi: Yes totally! I will change it.

apis/workloads/v1alpha1/lifecycle_recorder.go
ttamimi: Yes here is a live example that was generated in lab:
https://gocardless.atlassian.net/browse/TCSR-290
The app container is in waiting state.

ttamimi: Well, not exactly a crashbackoff loop. It's an imagepullbackoff loop.

ttamimi: Nice! Thanks for checking 😊

theobarberbany (approved):

PR #280, opened 2023-06-13
Update Kubernetes 1.24 and kind v0.19.0

rnaveiras:
Ensure that we test all the components against Kubernetes 1.24. We should aim for Kubernetes 1.27 in the future, but this is a good step forward as many behaviour changes are related to this version.

https://kubernetes.io/blog/2022/05/03/kubernetes-1-24-release-announcement/

In particular, we are looking at one change that directly impacts Vault. Service Account Tokens in Kubernetes v1.24, non-expiring service account tokens are no longer auto-generated.

You can read further details here:
https://eng.d2iq.com/blog/service-account-tokens-in-kubernetes-v1.24/

## How come this is working? without additional changes?

With any vault version before 1.9, this should break all the integration tests because the jwt iss validation will fail. 

You can read more details about this here https://developer.hashicorp.com/vault/docs/auth/kubernetes#kubernetes-1-21

You need to enable `disable_iss_validation=true` on those versions for the `auth/kubernetes/config` configuration.

From Vault 1.9.0, `disable_iss_validation` and `issuer` are deprecated, and the default for `disable_iss_validation` has changed to `true` for new Kubernetes auth mounts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,9 +136,9 @@ $ # install setup-envtest which configures etcd and kube-apiserver binaries for 
 $ # https://book.kubebuilder.io/reference/envtest.html#configuring-envtest-for-integration-tests
 $ # https://github.com/kubernetes-sigs/controller-runtime/tree/master/tools/setup-envtest#envtest-binaries-manager
 $ go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
-$ # configure envtest to use k8s 1.22.x binaries
-$ setup-envtest use -p path 1.22.x
-$ source <(setup-envtest use -i -p env 1.22.x)
+$ # configure envtest to use k8s 1.24.x binaries
+$ setup-envtest use -p path 1.24.x
+$ source <(setup-envtest use -i -p env 1.24.x)
 ```
 
 - **Unit**: Standard unit tests, used to exhaustively specify the functionality of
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -136,9 +136,9 @@ $ # install setup-envtest which configures etcd and kube-apiserver binaries for
 $ # https://book.kubebuilder.io/reference/envtest.html#configuring-envtest-for-integration-tests
 $ # https://github.com/kubernetes-sigs/controller-runtime/tree/master/tools/setup-envtest#envtest-binaries-manager
 $ go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
-$ # configure envtest to use k8s 1.22.x binaries
-$ setup-envtest use -p path 1.22.x
-$ source <(setup-envtest use -i -p env 1.22.x)
+$ # configure envtest to use k8s 1.24.x binaries
+$ setup-envtest use -p path 1.24.x
+$ source <(setup-envtest use -i -p env 1.24.x)
```
DragosDumitrache, line 141: Not a big deal, but I'd love it if this weren't so bash centric 🥺

DragosDumitrache (approved):

README.md
rnaveiras: haha, I feel the same way. let's see what I can do in the following PR's

PR #299, opened 2023-06-15
Update golang to 1.20.5 & Makefile improvements

rnaveiras:
- Golang 1.20.5

Used the last version of Golang, updating the github workflows and the
Dockerfile.

It also includes an update on the Ubuntu version.

- Update Makefile with new targets

Aim to simplify the user experience while developing locally.

Removed the GOARCH flag, so you're using your default ARCH by default. I
I think this will be simpler for everyone and reduce the number of binaries
to build.

Added helpers to run the acceptance tests; maybe we can also use them in
GitHub actions. Including makefile targets to install all tools required.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,56 +11,72 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
-.PHONY: build build-darwin build-linux build-all test generate manifests deploy clean docker-build docker-pull docker-push docker-tag controller-gen
+.PHONY: build build-darwin build-linux build-all clean fmt vet test acceptance-e2e acceptance-run acceptance-prepare acceptance-destory generate manifests install-tools deploy
 
 build: $(PROG)
-build-darwin: $(PROG:=.darwin_amd64)
-build-linux: $(PROG:=.linux_amd64)
+build-darwin: $(PROG:=.darwin)
+build-linux: $(PROG:=.linux)
 build-all: build-darwin build-linux
 
 # Specific linux build target, making it easy to work with the docker acceptance
-# tests on OSX
-bin/%.linux_amd64:
-	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 $(BUILD_COMMAND) -a -o $@ ./cmd/$*/.
+# tests on OSX. It uses by default your local arch
+bin/%.linux:
+	CGO_ENABLED=0 GOOS=linux $(BUILD_COMMAND) -a -o $@ ./cmd/$*/.
 
-bin/%.darwin_amd64:
-	CGO_ENABLED=0 GOOS=darwin GOARCH=amd64 $(BUILD_COMMAND) -a -o $@ ./cmd/$*/.
+bin/%.darwin:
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -11,57 +11,78 @@ else
...
 
 manifests: generate
 	$(CONTROLLER_GEN) crd paths="./apis/rbac/..." output:crd:artifacts:config=config/base/crds
 	$(CONTROLLER_GEN) crd paths="./apis/workloads/..." output:crd:artifacts:config=config/base/crds
 
+# See https://github.com/kubernetes-sigs/controller-runtime/tree/main/tools/setup-envtest
+install-tools:
+	go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.10.0
+	go install github.com/onsi/ginkgo/ginkgo@v1.16.5
+	go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
```
DragosDumitrache, line 68: Can we make this a dependency for any target that requires these tools?

Makefile
```diff
@@ -11,57 +11,78 @@ else
...
+
+install-tools-kubebuilder:
+	sudo mkdir /usr/local/kubebuilder
+	curl -fsL "https://github.com/kubernetes-sigs/kubebuilder/releases/download/v2.3.1/kubebuilder_2.3.1_$(go env GOOS)_$(go env GOARCH).tar.gz" | \
+		sudo tar -xvz --strip=1 -C /usr/local/kubebuilder
+	echo "you need to export KUBEBUILDER_ASSETS=/usr/local/kubebuilder/bin in your shell"
+
+install-tools-homebrew:
+	brew install kubernetes-cli kustomize kind
+	echo "you also need docker and go in your developer environment"
```
DragosDumitrache, line 78: Similarly for these

Makefile
```diff
@@ -11,57 +11,78 @@ else
...
 	kustomize build config/base | kubectl apply -f -
 
 deploy-production:
 	kustomize build config/overlays/production | kubectl apply -f -
 
-clean:
-	rm -rvf $(PROG) $(PROG:%=%.linux_amd64) $(PROG:%=%.darwin_amd64)
-
 docker-build:
 	docker build -t $(IMAGE):latest .
```
DragosDumitrache, line 87: This failed to build

Makefile
DragosDumitrache: Nvm, this is already the case. I only mentioned this because I failed to find gingko locally

Makefile
rnaveiras: I would prefer no make homebrew a dependency because it enforces how everyone configures our local development. For example, I don't use homebrew for any of these tools. I just added it to make it easier for people that do. I mentioned this target in the README.

Makefile
rnaveiras: looking

Makefile
rnaveiras: I think i missed the make test, fixing

PR #323, opened 2023-08-29
dockerfile: Split dependencies stage

benwh:
Make the build slightly quicker, by having a stage which is cached if our dependencies haven't changed.


benwh:
Opened this with the wrong source branch, so it now looks quite broken. Can't change the branch so closing, will reopen.

PR #335, opened 2024-08-19
Fix CI dependencies and Kustomize deprecations

benwh:
Here's a couple of fixes that we've got in our fork. Details in commit messages.

### DIFF
--- a/config/base/kustomization.yaml
+++ b/config/base/kustomization.yaml
@@ -5,8 +5,10 @@ kind: Kustomization
 namespace: theatre-system
 namePrefix: theatre-
 
-commonLabels:
-  app: theatre
+labels:
+- includeSelectors: true
+  pairs:
+    app: theatre
 
 resources:
   - crds/rbac.crd.gocardless.com_directoryrolebindings.yaml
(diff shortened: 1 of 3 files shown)


0x0013:
Hi @benwh, thanks for this PR.

It seems though that integration test fails after this change. From what I can see, since we are no longer defining the environment variable `THEATRE_IMAGE`, but rather replace it directly, the [vault manager overlay](https://github.com/gocardless/theatre/blob/62d067d69ad875ec5055ee610d681841f627688f/config/acceptance/overlays/vault-manager.yaml) used for acceptance test, which still overlays the command args with the env variable, now results in an invalid `read-a-secret-*` container, as the `image:` value no longer corresponds to an existing environment variable. 

We still need to define the `--theatre-image=` in the overlay, otherwise it will inherit the (production) value from the base, so it can't just be removed from the overlay.

I don't think reading the `THEATRE_IMAGE` value is the purpose of the tests, so it should be fine to directly replace it in the acceptance overlay with the test image used:

```yaml
# ...
          args:
            - --theatre-image=theatre:latest
            - --metrics-address=0.0.0.0
# ...
```

I see [the tests are failing](https://github.com/incident-io/theatre/actions/runs/10456147951/job/28952800423) in your fork as well, though the logs have timed out, I assume the reason is the same.

0x0013:
@benwh also note that we've merged #338 to update Go version, and with it, we've pinned envtest to 0.19, thus 5975be92d6e96e84d6d80156a7e9a466da61e4d5 would be unneeded as it pins to an earlier version.

benwh:
@0x0013 thanks for the pointers!

I've rebased to drop the superseded commit, and updated the acceptance kustomization to fix what I suspect are the issues there (I've been somewhat ignoring the `vault-manager` stuff, because we don't use that here)

config/base/kustomization.yaml
```diff
@@ -22,16 +24,27 @@ resources:
   - rbac/leader-election.yaml
   - cert-manager/certificate.yaml
 
-vars:
+replacements:
```
0x0013, line 27: with the new approach of repeating this replacement in overlay kustomization, there is a slight risk of someone modifying this replacement code, but forgetting to mirror the modification in overlay, in which case the replacement could be buggy, and the tests would still pass.

I don't think there's an elegant way to solve this, but could you add a code comment that the replacement code needs to be mirrored in `config/acceptance/kustomization.yaml`?

PR #336, opened 2024-11-07
Make reason mandatory when interactive

doomholderz:
Making reason entry mandatory on interactive console creations


benwh:
While I'm no longer a maintainer on this repo, I can provide some historical context:

The `runner` package in this repo was designed to be relatively un-opinionated, mostly as an inspiration and something that you could use when _developing_ against `theatre`. At GoCardless (and incident.io) we imported this package into our _own_ CLI, and only in _those_ CLI's would we encode organisation-specific logic like requiring a reason.

I don't think it's practical for anyone to be using the `runner` CLI here for their own organisation, given that it's not packaged and distributed nicely, hence I'd be against putting org-specific requirements into it.

PR #491, opened 2026-03-19
feat: more rollback metrics

IsaacJames:
### What:

Adds the following prometheus metrics:

- `theatre_rollback_completion_duration_seconds`: histogram to track time from rollback creation to terminal completion.
- `theatre_rollback_retry_count`: histogram to track retries per rollback (`attempts - 1`).

### DIFF
--- a/internal/controller/deploy/metrics.go
+++ b/internal/controller/deploy/metrics.go
@@ -49,8 +49,8 @@ var (
 )
 
 // buildRollbackLabels creates a prometheus.Labels map from the rollback's labels
-// plus the required namespace and status labels. Only includes labels that are
-// defined in rollbackLabelKeys to avoid Prometheus errors.
+// plus the required namespace and status labels. All label keys from rollbackLabelKeys
+// must be present to satisfy Prometheus cardinality requirements.
 func buildRollbackLabels(rollback *deployv1alpha1.Rollback, status string) prometheus.Labels {
 	labels := prometheus.Labels{
 		"status":      status,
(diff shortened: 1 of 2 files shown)


internal/controller/deploy/rollback_controller.go
```diff
@@ -26,11 +26,15 @@ const (
 	RequeueAfter = 15 * time.Second
 
 	// Maximum number of times to retry triggering a deployment
-	MaxRetryAttempts = 3
+	MaxRollbackAttempts = 3
```
IsaacJames, line 29: Drive-by: realised this naming is technically inaccurate, if you look at the usage, this is the maximum number of allowed _attempts_ not retries:
```
if rollback.Status.AttemptCount >= MaxRollbackAttempts {
	logger.Info("max rollback attempts exceeded", "attempts", rollback.Status.AttemptCount)
	return r.markRollbackFailed(ctx, logger, rollback, "max rollback attempts exceeded")
}
```

internal/controller/deploy/metrics.go
```diff
@@ -28,6 +28,24 @@ var (
...
+			Buckets: prometheus.DefBuckets,
+		},
+		rollbackLabelKeys,
+	)
+
+	rollbackRetryCount = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "theatre_rollback_retry_count",
+			Help:    "Number of retries performed by rollbacks before reaching terminal state",
+			Buckets: []float64{0, 1, 2, 3, 4, 5},
```
goelozev, line 45: Do we expect any rollbacks to fall in the `4` and `5` buckets, as our current rollback attempts are set to 3?

goelozev (approved):

internal/controller/deploy/metrics.go
IsaacJames: Claude suggested these buckets and I asked it the same question - it suggested a buffer in case the value gets increased in future, but perhaps it's not necessary. What do you think?

goelozev: Yeah, I assumed it's a buffer. I don't think it will cause confusion with people reading the metric tbf.

PR #498, opened 2026-04-20
Adding Claude Code review for theatre

goelozev:
Adding Claude-related files to enable reviews and better suggestions when using Claude to contribute to this repository.

### DIFF
--- a/.claude/agents/code-reviewer.md
+++ b/.claude/agents/code-reviewer.md
@@ -1,0 +1,84 @@
+---
+name: code-reviewer
+description: Expert code review specialist for the Theatre Kubernetes extensions project. Reviews Go code, CRD definitions, controllers, webhooks, and Kubernetes manifests for quality, security, correctness, and adherence to project conventions.
+tools: Read, Grep, Glob, Bash
+model: inherit
+---
+
+You are a senior Go and Kubernetes engineer performing code reviews on the Theatre project — GoCardless' Kubernetes extensions repository (`github.com/gocardless/theatre/v5`).
+
+## Project Context
+
+Theatre provides Kubernetes operators & admission controller webhooks. Read the [CLAUDE.md](../CLAUDE.md) first
+to understand the full context of the project.
+
+## When Invoked
+
+1. Run `git diff HEAD~1` (or the diff provided in context) to identify changed files
+2. Focus review on modified Go files, CRD types, controllers, webhooks, manifests, and tests
+3. Begin the review immediately without preamble
+
+## Review Checklist
+
+### Go Code Quality
+
+- Code is clear, idiomatic Go — follows standard patterns used elsewhere in the codebase
+- Functions and variables are well-named and scoped appropriately
+- No duplicated logic; shared helpers belong in `pkg/`
+- Proper use of Go error handling (no swallowed errors, errors wrapped with context)
+- Interfaces used appropriately; avoid over-abstraction
+- Concurrency is safe (race-free); check mutex usage and goroutine lifecycle
+
+### Kubernetes / Controller Patterns
+
+- Controllers follow the reconcile loop pattern correctly (idempotent, returns `ctrl.Result`)
+- Status conditions updated correctly; avoid patching the full object when a status patch suffices
+- RBAC markers (`+kubebuilder:rbac:...`) present and correct for new resource access
+- Webhook handlers validate inputs and return informative `admission.Denied` / `admission.Errored` responses
+- CRD type changes include updated `+kubebuilder:validation:` markers where appropriate
+- No direct `pods/exec` permissions granted unnecessarily (security-sensitive in this codebase)
+- The `pkg/recutil` package is used across the board to standardise reconciliation patterns and event emission
+- Follow the good practice guides in https://kubebuilder.io/reference/good-practices.html and https://github.com/kubernetes-sigs/controller-runtime/blob/main/FAQ.md. You might open any links in those pages, but do not open exceed going deeper than 2 levels.
+
+### Security
+
+- No secrets, API keys, or credentials hardcoded or logged
+- Vault integration: secrets fetched at runtime, not embedded in images or manifests
+- Admission webhooks fail closed (deny on error) where appropriate
+- RBAC minimal-privilege: roles grant only required verbs/resources
+
+### Testing
+
+- New behaviour is covered by tests at the appropriate level (unit → integration → acceptance)
+- Ginkgo tests use `Describe`/`Context`/`It` structure consistent with existing suites
+- Integration tests use `envtest`; acceptance tests use the Kind cluster via `cmd/acceptance`
+- No tests deleted or weakened without explicit justification
+
+### Manifests & Kustomize
+
+- CRD changes regenerated via `make manifests generate`
+- Kustomize overlays reference versioned bases (`?ref=vX.Y.Z`)
+- New CRDs included in both `config/crd` and referenced in `config/base` if needed
+
+## Output format
+
+```
+## Code Review: <PR title or branch>
+
+### 🔴 Critical Issues
+- [file:line] <issue> — <why it matters and what to do>
+
+### 🟡 Major Concerns
+- [file:line] <issue> — <why it matters and what to do>
+
+### 🔵 Minor Improvements
+- [file:line] <suggestion> — <rationale>
+
+### ✅ Positive Observations
+- <noteworthy good practices>
+
+### Summary & Next Steps
+<1-3 sentence overall assessment and clear recommended actions>
+```
+
+Only include sections that have content. Always include file and line number. Explain _why_ each issue matters, not just what it is.
--- a/.claude/agents/code-reviewer.md
+++ b/.claude/agents/code-reviewer.md
@@ -12,9 +12,17 @@ You are a senior Go and Kubernetes engineer performing code reviews on the Theat
 Theatre provides Kubernetes operators & admission controller webhooks. Read the [CLAUDE.md](../CLAUDE.md) first
 to understand the full context of the project.
 
+## How to get the diff
+
+- If given a PR number: run `gh pr diff <number>` and `gh pr view <number>` for title/description context
+- If given a branch name: `git diff <base_branch>...<head_branch>`
+- Otherwise: `git diff <base_branch>...HEAD`
+
+After getting the diff, identify all changed files and **read each one in full** to understand surrounding context before starting the review.
+
 ## When Invoked
 
-1. Run `git diff HEAD~1` (or the diff provided in context) to identify changed files
+1. Get the diff
 2. Focus review on modified Go files, CRD types, controllers, webhooks, manifests, and tests
 3. Begin the review immediately without preamble
 
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -1,0 +1,41 @@
+name: Claude Code
+
+on:
+  issue_comment:
+    types: [created]
+  pull_request_review_comment:
+    types: [created]
+  issues:
+    types: [opened, assigned]
+  pull_request_review:
+    types: [submitted]
+
+jobs:
+  claude:
+    if: |
+      (github.event.comment.author_association == 'MEMBER' || github.event.comment.author_association == 'COLLABORATOR') &&
+      (
+        (github.event_name == 'issue_comment' && contains(github.event.comment.body, '@claude')) ||
+        (github.event_name == 'pull_request_review_comment' && contains(github.event.comment.body, '@claude')) ||
+        (github.event_name == 'pull_request_review' && contains(github.event.review.body, '@claude')) ||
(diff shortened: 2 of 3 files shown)


.github/workflows/claude-code-review.yml
```diff
@@ -0,0 +1,38 @@
...
+  pull_request_review_comment:
+    types: [created]
+  issues:
+    types: [opened, assigned]
+  pull_request_review:
+    types: [submitted]
+
+jobs:
+  claude:
+    if: |
```
IsaacJames, line 15: I think we might want to add something like this to scope invocation to GC org members and direct collaborators only:
```suggestion
if: |
  (github.event.comment.author_association == 'MEMBER' ||
  github.event.comment.author_association == 'COLLABORATOR') &&
```

CLAUDE.md
```diff
@@ -0,0 +1,90 @@
...
+## Local Development Cluster (Kind)
+
+```shell
+make build
+make test
+make acceptance-e2e
+# or step-by-step:
+go run cmd/acceptance/main.go prepare  # provisions Kind cluster
+go run cmd/acceptance/main.go run --verbose
+go run cmd/acceptance/main.go destroy
```
IsaacJames:
```suggestion
make acceptance-prepare  # provisions Kind cluster
make acceptance-run
make acceptance-destroy
```

CLAUDE.md
```diff
@@ -0,0 +1,90 @@
...
+```shell
+make install-tools
+eval $(setup-envtest use -i -p env 1.24.x)
+```
+
+Three test levels:
+
+- **Unit** — `make test` (fast, no cluster needed)
+- **Integration** — `make test` (uses `envtest` with a temporary API server, no nodes)
+- **Acceptance** — `make acceptance-run` (full Kind cluster, slow, used sparingly)
```
IsaacJames: `make acceptance-run` will not work without first having run `make acceptance-prepare`, I think we should change this to `make acceptance-e2e`.

CLAUDE.md
```diff
@@ -0,0 +1,90 @@
...
+make build
+make test
+make acceptance-e2e
+# or step-by-step:
+go run cmd/acceptance/main.go prepare  # provisions Kind cluster
+go run cmd/acceptance/main.go run --verbose
+go run cmd/acceptance/main.go destroy
+```
+
+Re-run `prepare` after any code changes to rebuild and redeploy images.
```
IsaacJames:
```suggestion
Re-run `make acceptance-prepare` after any code changes to rebuild and redeploy images.
```

.claude/agents/code-reviewer.md
```diff
@@ -0,0 +1,95 @@
...
+- Code is clear, idiomatic Go — follows standard patterns used elsewhere in the codebase
+- Functions and variables are well-named and scoped appropriately
+- No duplicated logic; shared helpers belong in `pkg/`
+- Proper use of Go error handling (no swallowed errors, errors wrapped with context)
+- Interfaces used appropriately; avoid over-abstraction
+- Concurrency is safe (race-free); check mutex usage and goroutine lifecycle
+
+### Kubernetes / Controller Patterns
+
+- Controllers follow the reconcile loop pattern correctly (idempotent, returns `ctrl.Result`)
```
IsaacJames, line 34: do we want to mention the `recutil` package here?

.claude/agents/code-reviewer.md
```diff
@@ -0,0 +1,95 @@
...
+- Status conditions updated correctly; avoid patching the full object when a status patch suffices
+- RBAC markers (`+kubebuilder:rbac:...`) present and correct for new resource access
+- Webhook handlers validate inputs and return informative `admission.Denied` / `admission.Errored` responses
+- CRD type changes include updated `+kubebuilder:validation:` markers where appropriate
+- No direct `pods/exec` permissions granted unnecessarily (security-sensitive in this codebase)
```
IsaacJames, line 39: Adding some patterns from https://kubebuilder.io/reference/good-practices.html and https://github.com/kubernetes-sigs/controller-runtime/blob/main/FAQ.md:
```suggestion
- No direct `pods/exec` permissions granted unnecessarily (security-sensitive in this codebase)
- Avoid a design solution where the same controller reconciles more than one Kind
- Controller reconciliation loops should be idempotent, avoid writing reconciliation logic according to specific events
```

IsaacJames (commented):
Some suggestions for the context files

.github/workflows/claude-code-review.yml
IsaacJames: Looking at https://github.com/anthropics/claude-code-action/blob/main/docs/security.md I think this restriction is indirectly enforced via the action, but still worth adding IMO to deny more explicitly and save on runner mins!

CLAUDE.md
```diff
@@ -0,0 +1,90 @@
...
+
+- Go module path changed: `github.com/gocardless/theatre/v4` → `github.com/gocardless/theatre/v5`
+- `apis/` renamed to `api/`
+- Controllers/webhooks moved to `internal/` (no longer importable externally)
+- CRDs must now be imported separately in kustomizations:
+  ```yaml
+  resources:
+    - github.com/gocardless/theatre/config/crd?ref=v5.x.x
+    - github.com/gocardless/theatre/config/base?ref=v5.x.x
+  ```
```
0x0013: I don't see the usefulness of including this in agents file. It's very specific knowledge related to the upgrade, that is also already included in the readme.

CLAUDE.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Toolchain
+
+- **Go** `1.24.5`
+- **controller-gen** `v0.19.0` — CRD/webhook/RBAC manifest generation
+- **kustomize** `v5.7.1`
+- **golangci-lint** `v2.4.0`
+- **ginkgo** `v1.16.5`
+- **setup-envtest** — manages `etcd`/`kube-apiserver` binaries for integration tests
+- **kind** — Kubernetes-in-Docker for acceptance tests
```
0x0013, line 76: Not sure how useful this whole section is, but at least the specific versions should be removed, as those are prone to go out of date very quickly.

CLAUDE.md
0x0013, line 1: I assume this file was generated with LLM? It has been a bit controversial whether generated agents files help, or rather confuse, the models.

The rule of thumb seems to be to not include things that the agent can easily discover, but rather include mitigations after observing an agent arriving at a poor choice, to steer the model towards a better approach next time.

While it's hard for me to judge without doing actual evaluations, at the very least, we should refrain of including anything that is prone to becoming outdated quickly, such as specific versions.

.claude/agents/code-reviewer.md
0x0013: I wonder if it would be more useful to point the model at the good-practices URL, instead of copy-pasting select content in the prompt?

.claude/agents/code-reviewer.md
```diff
@@ -0,0 +1,95 @@
...
+- `cmd/` — CLI entry points (`rbac-manager`, `vault-manager`, `workloads-manager`, `theatre-consoles`, `theatre-secrets`)
+- `config/` — Kustomize manifests and CRD bases
+
+Key API groups under `*.crd.gocardless.com`:
+
+- **RBAC** — `DirectoryRoleBinding` (provisions `RoleBinding`s from Google Groups)
+- **Workloads** — `Console` (temporary dedicated pods for operational tasks)
+- **Vault** — `secrets-injector` webhook (injects Vault secrets into pod environments)
+
+Tests use the **Ginkgo** framework with `-race -randomizeSuites -randomizeAllSpecs`. Three levels: unit, integration (envtest), acceptance (Kind cluster).
```
0x0013: Is this section not fully duplicating content already in `CLAUDE.md`?

.claude/agents/code-reviewer.md
IsaacJames: I thought about this, but wondered if it would be bloating the context a bit (those pages also link out to a bunch of other pages). Could still be worth doing though!

.github/workflows/claude-code-review.yml
goelozev: Done!

CLAUDE.md
goelozev: Updated!

CLAUDE.md
goelozev: Updated!

.claude/agents/code-reviewer.md
goelozev: Good idea, I've added it!

CLAUDE.md
goelozev: Yes, no need! I've removed it!

.claude/agents/code-reviewer.md
goelozev: mentioned it!

CLAUDE.md
goelozev: Fair enough, I've removed them!

CLAUDE.md
goelozev: Yes, these have been mainly generated using an LLM. I have a follow-up pull request where I would assess how aware it is in the context provided.

.claude/agents/code-reviewer.md
goelozev: Yes, instead of copying it, I've updated the prompt to read the `CLAUDE.md` file before performing the review!

0x0013: Hmm, I would assume Claude Reviewer would do that implicitly, but I don't have a reference to docs stating so.

CLAUDE.md
0x0013: > assess how aware it is in the context provided.

Yeah, that's a great idea! Ideally, we would eval different context inclusions, as well as no additional context (absent file), with the same task/prompt, and compare the output and spent token count.

.claude/agents/code-reviewer.md
```diff
@@ -0,0 +1,84 @@
...
+You are a senior Go and Kubernetes engineer performing code reviews on the Theatre project — GoCardless' Kubernetes extensions repository (`github.com/gocardless/theatre/v5`).
+
+## Project Context
+
+Theatre provides Kubernetes operators & admission controller webhooks. Read the [CLAUDE.md](../CLAUDE.md) first
+to understand the full context of the project.
+
+## When Invoked
+
+1. Run `git diff HEAD~1` (or the diff provided in context) to identify changed files
```
0x0013: I wonder if this is accurate. Initially, I thought it was a mistake, but I'm not sure what the exact state is in the CI check that would be running.

The reason I'm mentioning this is, that on a normal pull of the PR branch locally, `git diff HEAD~1` would only show the diff of the latest commit, not the whole PR, and we would want something like `git diff <default-branch>` instead.

However, if the final branch presented to the context of the CI environment, is a synthetic sqash-and-merge result, then `HEAD~1` would point to the current `HEAD` of default branch already, so it would be what we want.

I assume the behavior is the latter, but still wanted to point it out just in case.

If we're not sure, instructing the model in natural language would probably work better, such as "inspect the changes between this PR and the default branch" or similar - models are more than capable of generating git commands, and will probably know better what command to invoke specifically, e.g. maybe formatting the output in a non-human friendly format, inspecting individual commits, etc.

0x0013 (dismissed):
I'm approving, after reviewing the responses to my prior review, though I have added an additional comment.

I did not review the changes related to Isaac's prior review.

.claude/agents/code-reviewer.md
goelozev: You're right. I've copied the instruction from an agent from another repository, which I verified works.

.claude/agents/code-reviewer.md
goelozev: The Claude Code Reviewer is a different product from the GitHub Action we are using. I couldn't find a definitive documentation stating that the action would read the `CLAUDE.md` file, so I made sure to instruct the agent.

.claude/agents/code-reviewer.md
```diff
@@ -0,0 +1,92 @@
...
+You are a senior Go and Kubernetes engineer performing code reviews on the Theatre project — GoCardless' Kubernetes extensions repository (`github.com/gocardless/theatre/v5`).
+
+## Project Context
+
+Theatre provides Kubernetes operators & admission controller webhooks. Read the [CLAUDE.md](../CLAUDE.md) first
+to understand the full context of the project.
+
+## How to get the diff
+
+- If given a PR number: run `gh pr diff <number>` and `gh pr view <number>` for title/description context
```
0x0013, line 17: I wonder - do we have `gh` available and set up in the CI environment? Or maybe it's irrelevant, since this is not exactly "given a PR number" case..

Technically, we could do without the different approaches of getting a diff, given that we expect the CI environment to already have checked out the PR branch (with an optional merge at the end of it, depending on settings)

However, it's probably worth leaving it in, given that the agent could be instantiated in a different environment, than the CI.

0x0013 (approved):

PR #499, opened 2026-04-20
Rollbacks: Adding ArgoCD deployer backend

goelozev:
### Summary

Adding ArgoCD CI/CD backend for the rollback controller. This implementation will allow for faster rollbacks as it will skip our current CI system and directly perform the rollback. To run the rollback controller with ArgoCD, you will need to provide the following cmd flags:
- `--cicd-backend=argocd`
- `--argocd-server-url=<your-argocd-url>` the base URL of your ArgoCD server (e.g. `https://argocd.yourcompany.dev`)
- `--argocd-auth-token=<your-argocd-token>` authentication token for a robot account, defined in `argocd-cm (e.g. `accounts.theatre: apiKey`)
- `--argocd-app-name-template="<template>"` - define template to resolve the application name. In GC, we are using the `compute-lab-{{.Namespace}}-{{.Target}}` template for all ArgoCD application names. If you wish, you can modify that template, but currently, the only available fields you can use are `.Namespace` and `.Target`. You can also override this on a Rollback level by providing the `argocd_app_name` deployment option.

#### Requirements
Rollbacks using this CI/CD backend requires the following deployment options:
- `target_revision` - a specific revision or a JSONPath in the Release to find the revision (e.g. `'{.config.revisions[?(@.name=="infrastructure")].id}'`).

Optional deployment options:
- `app_revision` - a specific revision or a JSONPath in the Release to find the revision (e.g. `'{.config.revisions[?(@.name=="application")].id}'`).
- `argocd_app_name` - you can provide this deployment option if you wish to specify a specific application name if the application is not using the template provided in `--argocd-app-name-template`.
- `argocd_add_sync_window` - when set to true, a sync window will be added to the application, which will pause any further deployments

#### Changes to the Deployer interface
The Deployer interface (used to provide structure for CI/CD backends) now has a `PostDeploymentHooks` method, which will be called after the Rollback is deemed Healthy.

#### Tests
Tests conducted in our lab environment:
1. Rollback with `target_revision`, `app_revision` and `argocd_add_sync_window=true` set ✅ 
2. Rollback with `target_revision` and `argocd_add_sync_window=true` set ✅ 
3. Rollback with `target_revision` only set ✅


internal/controller/deploy/rollback_controller.go
```diff
@@ -258,7 +258,14 @@ func (r *RollbackReconciler) pollDeploymentStatus(ctx context.Context, logger lo
...
 	switch statusResp.Status {
 	case cicd.DeploymentStatusSucceeded:
-		return r.markRollbackSucceeded(ctx, logger, rollback, statusResp.Message)
+		message := statusResp.Message
+		err := r.triggerPostDeploymentHooks(ctx, rollback, toRelease)
+		if err != nil {
+			logger.Error(err, "failed to trigger post deployment hooks")
+			message = fmt.Sprintf("%s (post-deployment hooks failed: %v)", message, err)
+		}
+
```
goelozev, line 267: The `PostDeploymentHooks` are currently non-retryable, as we don't have a field in the status object. I was thinking of using the `status.attemptCount` field, but it might introduce some confusion. @IsaacJames what's your opinion.

pkg/cicd/deployer.go
```diff
@@ -73,6 +73,10 @@ type Deployer interface {
 	// The deploymentID should be the ID returned from TriggerDeployment.
 	GetDeploymentStatus(ctx context.Context, deploymentID string) (*DeploymentResult, error)
 
+	// PostDeploymentHooks is called after a deployment is successful.
+	// E.g. adding a sync window on a ArgoCD application
```
IsaacJames:
```suggestion
	// E.g. adding a deny sync window on a ArgoCD application
```

pkg/cicd/argocd/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Deployment Options
+
+These are set per Rollback resource under `deploymentOptions`.
+
+### Required
+
+| Key               | Description                                                                                                                                                                     |
+| ----------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `target_revision` | The infrastructure revision to deploy. Can be a literal revision string or a JSONPath expression into the Release (e.g. `{.config.revisions[?(@.name=="infrastructure")].id}`). |
```
IsaacJames: Do we want to accept literal revision string here (or for `app_revision` )? I'm thinking if someone creates a `Rollback` to `release_A` and then specifies literal string revisions that aren't actually associated with that release, then the target state would have nothing to do with the target release, which would be misleading.

internal/controller/deploy/rollback_controller.go
IsaacJames: I think it would be confusing yes, and retries here are probably not necessary.

pkg/cicd/argocd/deployer.go
```diff
@@ -0,0 +1,398 @@
...
+			return cicd.DeploymentStatusFailed, op.Message
+		case OperationPhaseSucceeded:
+			return cicd.DeploymentStatusSucceeded, op.Message
+		default:
+			return cicd.DeploymentStatusPending, fmt.Sprintf("Operation phase: %s", op.Phase)
+		}
+	}
+
+	return cicd.DeploymentStatusPending, "No active operation"
+}
```
IsaacJames: I'm still unsure about mapping the operation status of whatever operation happens to be running to the deployment status. I'm not sure if there is an easy way to verify the operation is the same rollback  sync that the controller initiated, but not doing so could produce some weird behaviour. E.g. if the controller fails to trigger a sync, but some other actor (CICD, human etc.) triggers one, the rollback could get marked as in progress/completed.

IsaacJames: Also, I'm not sure if we need to verify whether the operation is even a sync operation (though I'm unsure if other types of operations exist as I write this).

0x0013:
While comparing different promp context inclusions for claude reviews, I was running the experiments on the changes in this PR. Pasting one of the reviews (which was considered the highest quality review)

Note that I haven't personally verified the validity of the review.

## Code Review: `rollback-argocd-backend` vs `pre-release-5.1.0`

This PR introduces a new ArgoCD CI/CD backend for the rollback controller, adds a `PostDeploymentHooks` method to the `Deployer` interface, and wires everything up through the rollback controller and `main.go`. The overall design is solid and the test coverage is good, but there are several correctness bugs that must be addressed before merge.

---

### 🔴 Critical Issues

- **[`pkg/cicd/argocd/deployer.go:161`] — nil pointer panic in `resolveAppName` when no template is configured**
  `NewDeployer` accepts an empty `appNameTmpl` and leaves `d.appNameTemplate` as `nil`. If any `Rollback` doesn't include `argocd_app_name` in its options and the deployer was started without `--argocd-app-name-template`, `d.appNameTemplate.Execute(...)` will panic at runtime. The README acknowledges this is valid ("not required if every Rollback provides `argocd_app_name`"), but this should return a clean error rather than crash the controller process. Fix: add a nil-guard before the `.Execute` call:
  ```go
  if d.appNameTemplate == nil {
      return "", fmt.Errorf("no app name template configured and %q not provided in options", AppNameKey)
  }
  ```

- **[`pkg/cicd/argocd/types.go:76`] — missing `omitempty` on `Plugin` field silently clears ArgoCD plugin config**
  `applicationPatchSource.Plugin` has no `omitempty` tag and is not a pointer:
  ```go
  type applicationPatchSource struct {
      TargetRevision string                 `json:"targetRevision"`
      Plugin         applicationPatchPlugin `json:"plugin"`  // ← no omitempty
  }
  ```
  When `appRevision` is empty, `Plugin` is the zero value and serialises to `"plugin":{"env":null}`. A JSON merge-patch interprets this as an explicit instruction to set the plugin env to null, **clearing any existing ArgoCD plugin configuration on the application**. Fix: make `Plugin` a pointer with `omitempty`:
  ```go
  Plugin *applicationPatchPlugin `json:"plugin,omitempty"`
  ```

- **[`pkg/cicd/argocd/deployer.go:295–330`] — `addSyncWindow` issues a destructive full-replace PUT, silently erasing unmapped project fields**
  The flow is GET → decode into `projectResponse` → mutate → PUT back. Go's JSON decoder discards any project fields not defined in `projectSpec` (e.g., `description`, `orphanedResources`, `signatureKeys`, `namespaceResourceWhitelist`, `permitOnlyProjectScopedClusters`, etc.). ArgoCD's `PUT /api/v1/projects/{name}` is a full replace, not a merge, so all those fields will be **permanently cleared** from the project on the first successful hook call. This is a serious data-loss risk for any project with non-trivial configuration. Fix: either use ArgoCD's dedicated sync-window API endpoints (`POST /api/v1/projects/{project}/syncwindows`), or capture the raw JSON response and manipulate it without full round-tripping through a partial struct.

---

### 🟡 Major Concerns

- **[`pkg/cicd/argocd/deployer.go:175–202`] — `findRevision` is dead code**
  The function is defined but never called anywhere in the codebase. The deployer takes `target_revision` and `app_revision` directly as string options and never inspects `req.ToRelease.Revisions`. Either wire it into the flow (providing the safety of named-revision resolution from the Release object) or remove it. Dead code in a new package is a maintenance hazard.

- **[`pkg/cicd/argocd/deployer.go:243`] — `revision` parameter in `syncApplication` is accepted but silently unused**
  ```go
  func (d *Deployer) syncApplication(ctx context.Context, appName, revision string) error {
      path := fmt.Sprintf("/api/v1/applications/%s/sync", ...)
      resp, err := d.doRequest(ctx, http.MethodPost, path, []byte{}, ...)
  ```
  The ArgoCD sync API accepts a `revision` in its request body. Passing it would pin the sync to the exact revision just set in `updateApplication`, eliminating a TOCTOU window where another controller could change `targetRevision` between the PATCH and the sync. At minimum, remove the unused parameter to avoid confusion.

- **[`pkg/cicd/argocd/deployer.go:309–315`] — `addSyncWindow` appends without deduplication; repeated controller reconciliations will accumulate duplicate sync windows**
  ```go
  proj.Spec.SyncWindows = append(proj.Spec.SyncWindows, SyncWindow{...})
  ```
  Every time `PostDeploymentHooks` is invoked (e.g., on controller restart, or if `GetDeploymentStatus` transiently returns `Succeeded` more than once), a new `deny *` sync window is appended to the ArgoCD project. Over time this can accumulate dozens of deny windows. Fix: check for an existing matching window before appending.

- **[`cmd/rollback-manager/main.go:152`] — `http.DefaultClient` (no timeout) used for ArgoCD**
  ```go
  return argocddeployer.NewDeployer(http.DefaultClient, ...)
  ```
  The Go `http.DefaultClient` has no request timeout. A hung ArgoCD API call will block the reconcile goroutine indefinitely, stalling rollback operations. The GitHub backend avoids this by using `golang.org/x/oauth2` which wraps a proper client. Fix: create a dedicated `&http.Client{Timeout: ...}` (e.g., 30s) rather than reusing the default.

- **[`internal/controller/deploy/rollback_controller.go:262–268`] — post-deployment hook failure silently results in a `Succeeded` rollback status**
  ```go
  err := r.triggerPostDeploymentHooks(ctx, rollback, toRelease)
  if err != nil {
      logger.Error(err, "failed to trigger post deployment hooks")
      message = fmt.Sprintf("%s (post-deployment hooks failed: %v)", message, err)
  }
  return r.markRollbackSucceeded(ctx, logger, rollback, message) // ← always succeeds
  ```
  Given that the primary hook is adding a **deny sync window** — a safety mechanism to prevent further auto-syncs after a rollback — silently succeeding when this fails could leave the cluster in a state where the rollback completed but ArgoCD will immediately re-sync to the "bad" revision. The current behaviour needs at least a dedicated status condition or metric to make hook failures observable; the contract ("hooks are best-effort") should also be explicit in the interface doc-comment.

- **[`internal/controller/deploy/rollback_controller.go:299`] — `ParseDeploymentOptions` error silently discarded (also at line 190)**
  ```go
  parsedOptions, _ := cicd.ParseDeploymentOptions(...)
  ```
  Both `triggerDeployment` and the new `triggerPostDeploymentHooks` discard the error. If a JSONPath expression is malformed, the key is silently dropped from `parsedOptions`. For the hooks path this means `argocd_add_sync_window` could silently be absent, causing the sync window to be skipped with only an `Info` log at best. This is a pre-existing pattern but it's now being extended to a new code path — worth fixing.

---

### 🔵 Minor Improvements

- **[`pkg/cicd/argocd/types.go:131`] — `applicationProjectSpec` is an unused exported-like type**
  ```go
  type applicationProjectSpec struct {
      SyncWindows []SyncWindow `json:"syncWindows"`
  }
  ```
  This type is defined but never referenced. Remove it to keep the package clean.

- **[`pkg/cicd/argocd/types.go:118`] — `SyncWindow` is exported while all other types in the file are unexported**
  There's no clear reason for `SyncWindow` to be exported (it's an internal ArgoCD API mirror type). Make it unexported (`syncWindow`) for consistency, unless it is intentionally part of the public API surface.

- **[`pkg/cicd/argocd/deployer.go:243`] — `[]byte{}` is not valid JSON for an empty POST body**
  ```go
  resp, err := d.doRequest(ctx, http.MethodPost, path, []byte{}, "application/json")
  ```
  An empty byte slice is not valid JSON. Pass `nil` (which `doRequest` correctly handles by using a nil reader) or `[]byte("{}")` if ArgoCD requires a JSON body.

- **[`pkg/cicd/deployer.go:77`] — grammar in interface doc-comment**
  `"E.g. adding a sync window on a ArgoCD application"` → `"an ArgoCD application"`. Also, "E.g." is typically followed by a comma in English usage.

- **[`pkg/cicd/argocd/deployer.go:119–135`] — `PostDeploymentHooks` re-resolves the app name redundantly**
  `PostDeploymentHooks` calls `resolveAppName` to get the app name and then uses it for `getApplication`. But `deploymentID` (the third argument) is already the ArgoCD app name as returned by `TriggerDeployment`. Using `deploymentID` directly for the `getApplication` call would be simpler and avoid re-running template logic.

- **[`pkg/cicd/argocd/README.md:50`] — hardcoded sync window parameters not documented**
  The `addSyncWindow` implementation always adds a `deny / * * * * * / 1h / applications: ["*"] / manualSync: false` window, but these values are not mentioned in the README. Users enabling `argocd_add_sync_window: true` should know what they'll get, especially that it applies to **all applications** in the project and blocks manual syncs.

---

### ✅ Positive Observations

- Good use of `var _ cicd.Deployer = &Deployer{}` compile-time interface assertion in `deployer.go`.
- Comprehensive test coverage using `gock` for HTTP mocking; all key error paths (404, 429, 500) are covered with retryability assertions.
- Clean `handleErrorResponse` helper avoids duplicating HTTP error handling logic across methods.
- `url.PathEscape` is correctly used for all path parameters to handle special characters in app/project names.
- The `BeforeEach`/`JustBeforeEach` pattern is used well in tests to separate setup from execution, keeping individual test cases focused.
- The `README.md` is clear and provides a helpful status-mapping table.

---

### Summary & Next Steps

The ArgoCD backend is well-structured and the test suite is thorough, but there are **three correctness bugs that must be fixed before merge**: the nil-template panic in `resolveAppName`, the missing `omitempty` on `Plugin` silently clearing ArgoCD plugin config, and the destructive full-replace PUT in `addSyncWindow` erasing unmapped project fields. The sync-window deduplication issue is also a production reliability concern that should be addressed. After those are resolved, the `http.DefaultClient` timeout and the silent hook-failure behaviour should be addressed to make the new code production-safe.

pkg/cicd/argocd/README.md
goelozev: Good point! I will update the documentation to not state that you can provide a literal revision.

> I'm thinking if someone creates a Rollback to release_A and then specifies literal string revisions that aren't actually associated with that release, then the target state would have nothing to do with the target release, which would be misleading.

With the interfaces we provide within GC, I don't think this will be an issue, but still a valid point. You can also make the case that people will be able to use a JSONPath that doesn't target the original `target_revision`, which will result in an inaccurate rollback.

pkg/cicd/argocd/deployer.go
goelozev: That's a valid concern!

After a bit of back and forth with Claude today, I've updated the logic so the application name and the set of revisions are recovered as the deployment URL. Then, when the `GetDeploymentStatus` is called, the logic makes sure that the latest sync operation has the target revisions we are rolling back to. Even if an external action (person or CD) syncs to the same revision, the rollback will be marked as successful.

Only if the external actor syncs to a different set of revisions before the controller reconciles the rollback to a terminal state, or the sync operation fails, will the rollback be marked as failed.

goelozev:
> While comparing different promp context inclusions for claude reviews, I was running the experiments on the changes in this PR. Pasting one of the reviews (which was considered the highest quality review)
> 
> Note that I haven't personally verified the validity of the review.

@0x0013 I've minimised your comment as I changed the logic of the argocd deployer, and I will run the Claude code review separately on the new logic.

goelozev:
@claude review this PR!

goelozev:
@claude review this PR!

goelozev:
> **Claude finished @goelozev's task in 3m 6s** —— [View job](https://github.com/gocardless/theatre/actions/runs/25550438756)
> ...
> ### 🔴 Critical Issues
> **[`pkg/cicd/argocd/types.go:101-104` + `deployer.go:219-234`] `Plugin` struct field with `omitempty` never omitted — may silently clear ArgoCD plugin config**
Addressed ✅ 


> ### 🟡 Major Concerns
> **[`cmd/rollback-manager/main.go:139`, `main.go:149`] Deployers use dev-mode logger instead of the application logger**

Addressed ✅ 

> **[`pkg/cicd/argocd/deployer.go:402-418`] Dead code: `mapSyncStatus` method is never called**

Addressed ✅ 

> **[`pkg/cicd/argocd/deployer.go:371-377`] Sync window is applied to the entire project (`Applications: ["*"]`), not just the specific application**

That's expected behaviour!

> **[`pkg/cicd/argocd/deployer.go:371-377`] Sync windows are never deduplicated**
Deduplication is handled by the ArgoCD API.

> **[`internal/controller/deploy/rollback_controller.go:262-268`] `PostDeploymentHooks` failures are silently swallowed**

This is expected behaviour!

> **[`pkg/cicd/argocd/deployer.go:367-369`, `deployer.go:383-386`] `addSyncWindow` returns plain errors — not `*cicd.DeployerError`**

Addressed ✅ 

> **[`pkg/cicd/argocd/deployer.go:341`] `getApplication` hardcodes `"GetDeploymentStatus"` as the operation name regardless of caller**

Addressed ✅ 

> When `getApplication` is called from `PostDeploymentHooks`, any resulting `DeployerError` will report `Operation: "GetDeploymentStatus"`, making diagnostics misleading. The operation name should be passed as a parameter.

Addressed ✅ 

> ### 🔵 Minor Improvements
> **[`pkg/cicd/argocd/README.md:42-51`] Status mapping table is inaccurate after refactor**
 
Addressed ✅ 

> **[`pkg/cicd/argocd/README.md`] `argocd_add_sync_window` option missing from documentation**

Addressed ✅ 

> **[`pkg/cicd/argocd/deployer.go:26`] `deploymentIDSep = "::"` fragility**

Not addressed.

> **[`pkg/cicd/argocd/deployer_test.go`] Missing test: `GetDeploymentStatus` when `operationState.Phase == "Succeeded", but `revisionMatches` is also false for the `appRevision` component**

Addressed ✅

PR #501, opened 2026-04-22
Adding Claude Code files to pre-release branch

goelozev:
bringing the changes from https://github.com/gocardless/theatre/pull/498, the first 2 commits are from that pull and also adding 2 README.md files for the deploy group CRDs and controllers

### DIFF
--- a/.claude/agents/code-reviewer.md
+++ b/.claude/agents/code-reviewer.md
@@ -1,0 +1,92 @@
+---
+name: code-reviewer
+description: Expert code review specialist for the Theatre Kubernetes extensions project. Reviews Go code, CRD definitions, controllers, webhooks, and Kubernetes manifests for quality, security, correctness, and adherence to project conventions.
+tools: Read, Grep, Glob, Bash
+model: inherit
+---
+
+You are a senior Go and Kubernetes engineer performing code reviews on the Theatre project — GoCardless' Kubernetes extensions repository (`github.com/gocardless/theatre/v5`).
+
+## Project Context
+
+Theatre provides Kubernetes operators & admission controller webhooks. Read the [CLAUDE.md](../CLAUDE.md) first
+to understand the full context of the project.
+
+## How to get the diff
+
+- If given a PR number: run `gh pr diff <number>` and `gh pr view <number>` for title/description context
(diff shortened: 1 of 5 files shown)


0x0013 (commented):
Looks good, however, could we pull in (cherry-pick or rebase) the merged commit from `master`, separately?

Though I don't expect these changes to cause conflicts, it would still IMO be better, as that commit will be exactly the same on both branches, and can thus be simply fast-forwarded when we merge pre-release into master.

goelozev:
> Looks good, however, could we pull in (cherry-pick or rebase) the merged commit from `master`, separately?

Good point, I've now done that!

IsaacJames (approved):

.claude/agents/code-reviewer.md
```diff
@@ -0,0 +1,92 @@
...
+tools: Read, Grep, Glob, Bash
+model: inherit
+---
+
+You are a senior Go and Kubernetes engineer performing code reviews on the Theatre project — GoCardless' Kubernetes extensions repository (`github.com/gocardless/theatre/v5`).
+
+## Project Context
+
+Theatre provides Kubernetes operators & admission controller webhooks. Read the [CLAUDE.md](../CLAUDE.md) first
+to understand the full context of the project.
```
0x0013: I'm not sure if I mentioned this on the `master` PR, but I'm quite sure this is not necessary - the purpose of `AGENTS.md` (or `CLAUDE.md` in this case) is that it gets injected in the context automatically.

Besides that, there are 2 other issues:
- the relative path points 1 directory up (`.claude/`), while the file is actually in the root, 2 parent dirs up
- the contents of a custom agent (code-reviewer) is the agent's verbatim prompt, not an instruction to read the file. Most likely, the agent's CWD will be the root of repo (where it gets launched from), so this will instruct it to look for the file one dir up, outside the repo.

goelozev: Yes, you did mention it previously, but I might have forgotten to action it - I've added it to this pull https://github.com/gocardless/theatre/pull/503.

goelozev: Done, I've cherry-picked it in this branch, too!

PR #505, opened 2026-06-18
Add eviction annotation for theatre-secrets-install

epatmalnieks-gc:
The theatre-secrets-install init container mounts an emptyDir volume to share secrets into the main container. The cluster autoscaler treats any pod with a local emptyDir volume as unsafe to evict, which can prevent nodes from being scaled down.

This PR introduces a change where the webhook now sets cluster-autoscaler.kubernetes.io/safe-to-evict-local-volumes: theatre-secrets-install on any pod it mutates, which signals the cluster autoscaler that it is safe to evict this volume.


0x0013 (approved):
LGTM

VSpike (approved):

PR #506, opened 2026-06-25
feat(rollback): add automated rollback e2e test

IsaacJames:
> [!NOTE]  
> I've added these tests under `rollback-manager`, even though they also implicitly test the release controller, however we don't have a shared manager for the three deploy controllers under `cmd`. Happy to take suggestions on this.

### What

Adds an acceptance/E2E test suite for automated rollbacks under `cmd/rollback-manager/acceptance`.

The new test verifies the full automated rollback flow across the release, automated rollback, and rollback controllers:

- creates an enabled `AutomatedRollbackPolicy`
- creates release analysis templates used by `release-manager`
- marks the previous release healthy via a successful health `AnalysisRun`
- fails rollback analysis for the active release via a failed `AnalysisRun`
- verifies `release-manager` sets `RollbackRequired=True`
- verifies `automated-rollback-manager` creates a `Rollback`
- verifies `rollback-manager`
- verifies the automated rollback policy is disabled after rollback completion

### DIFF
--- a/cmd/rollback-manager/acceptance/acceptance.go
+++ b/cmd/rollback-manager/acceptance/acceptance.go
@@ -63,7 +63,7 @@ func (r *Runner) Run(logger kitlog.Logger, config *rest.Config) {
 		})
 
 		Specify("Happy Path", func(ctx context.Context) {
-			By("Create a automated rollback policy")
+			By("Create an automated rollback policy")
 			targetName = generateName("target")
 			createPolicy(ctx, kubeClient, targetName, true)
 
@@ -126,7 +126,7 @@ func waitForRollbackWebhook(ctx context.Context, kubeClient client.Client, logge
 			return false
 		}
 		return true
-	}).WithContext(ctx).Should(Equal(true))
+	}).WithContext(ctx).Should(BeTrue())
 }
 
 func cleanupRollbackTestResources(ctx context.Context, kubeClient client.Client, targetName string) {
--- a/cmd/rollback-manager/acceptance/acceptance.go
+++ b/cmd/rollback-manager/acceptance/acceptance.go
@@ -46,14 +46,14 @@ func (r *Runner) Prepare(logger kitlog.Logger, config *rest.Config) error {
 }
 
 func (r *Runner) Run(logger kitlog.Logger, config *rest.Config) {
-	Describe("Automated Rollback", func() {
+	Describe("Automated Rollback", Ordered, func() {
 		var (
 			kubeClient         client.Client
 			targetName         string
 			previousTargetName string
 		)
 
-		BeforeEach(func() {
+		BeforeAll(func() {
 			kubeClient = newClient(config)
 			waitForRollbackWebhook(kubeClient, logger)
 		})
--- a/cmd/rollback-manager/acceptance/acceptance.go
+++ b/cmd/rollback-manager/acceptance/acceptance.go
@@ -64,7 +64,7 @@ func (r *Runner) Run(logger kitlog.Logger, config *rest.Config) {
 
 		Specify("Happy Path", func() {
 			By("Create a automated rollback policy")
-			targetName = generateTargetName()
+			targetName = generateName("target")
 			createPolicy(kubeClient, targetName, true)
 
 			By("Create rollback analysis")
@@ -160,10 +160,6 @@ func cleanupRollbackTestResources(kubeClient client.Client, targetName string) {
 
 func generateName(prefix string) string {
 	return fmt.Sprintf("%s-%d", prefix, testCounter.Add(1))
-}
-
-func generateTargetName() string {
-	return generateName("target")
 }
 
 func createReleaseWithLabels(kubeClient client.Client, targetName string, annotations, labels map[string]string) *deployv1alpha1.Release {
--- a/cmd/rollback-manager/acceptance/acceptance.go
+++ b/cmd/rollback-manager/acceptance/acceptance.go
@@ -69,12 +69,12 @@ func (r *Runner) Run(logger kitlog.Logger, config *rest.Config) {
 
 			By("Create rollback analysis")
 			previousTargetName = generateName("previous-target")
-			createAnalysisTemplate(kubeClient, targetName, previousTargetName, "health", "true")
-			createAnalysisTemplate(kubeClient, targetName, targetName, "rollback", "true")
+			createAnalysisTemplate(kubeClient, targetName, previousTargetName, "health")
+			createAnalysisTemplate(kubeClient, targetName, targetName, "rollback")
 
 			By("Create releases")
 			previousRelease := createActiveReleaseWithLabels(kubeClient, targetName, map[string]string{"target-name": previousTargetName})
-			previousAnalysisRun := expectAnalysisRunCreated(kubeClient, previousTargetName, "health", "true", targetName)
+			previousAnalysisRun := expectAnalysisRunCreated(kubeClient, previousTargetName, "health", targetName)
 			completeAnalysisRun(kubeClient, previousAnalysisRun.Name, analysisv1alpha1.AnalysisPhaseSuccessful)
 			expectReleaseHealthy(kubeClient, previousRelease.Name)
 			activeRelease := createActiveReleaseWithLabels(kubeClient, targetName, map[string]string{"target-name": targetName})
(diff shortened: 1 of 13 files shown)


cmd/rollback-manager/acceptance/acceptance.go
```diff
@@ -0,0 +1,326 @@
...
+}
+
+func (r *Runner) Run(logger kitlog.Logger, config *rest.Config) {
+	Describe("Automated Rollback", func() {
+		var kubeClient client.Client
+
+		BeforeEach(func() {
+			kubeClient = newClient(config)
+			waitForRollbackWebhook(kubeClient, logger)
+		})
```
0x0013: The use of `BeforeEach` seems a bit odd here:

1. currently, we only have a single test spec (happy path);
2. if/when we add additional tests, it seems more fitting to init the client once for all tests - maybe in `BeforeAll` block, if possible, or maybe in the `Prepare` function?
3. same for `waitForRollbackWebhook` - no need to poll for webhooks before every test.

I see that this is a pre-existing pattern used in other e2e acceptance tests, but I'm not convinced it is the logical approach.

cmd/rollback-manager/acceptance/acceptance.go
```diff
@@ -0,0 +1,326 @@
...
+
+		Specify("Happy Path", func() {
+			By("Create a automated rollback policy")
```
0x0013:
```suggestion
			By("Create an automated rollback policy")
```

cmd/rollback-manager/acceptance/acceptance.go
```diff
@@ -0,0 +1,326 @@
...
+		},
+		ReleaseConfig: deployv1alpha1.ReleaseConfig{
+			TargetName: targetName,
+			Revisions: []deployv1alpha1.Revision{
+				{Name: "application-revision", ID: generateName("app")},
+				{Name: "infrastructure-revision", ID: generateName("infra")},
+			},
+		},
+	}
+	Expect(kubeClient.Create(context.TODO(), release)).To(Succeed())
```
0x0013: It would be better to thread through Ginkgo's `SpecContext` ([docs](https://onsi.github.io/ginkgo/#spec-timeouts-and-interruptible-nodes)) instead of using `context.TODO()` - this is the canonical Ginkgo v2 pattern, and would cancel the context as appropriate, for example, cancelling inflight calls when a spec fails, handling spec timeouts etc.

This also applies to other places that pass context.

The prior e2e tests also use `context.TODO()`, so this would be an improvement, as opposed to alignment with our currently used patterns.

cmd/rollback-manager/acceptance/acceptance.go
```diff
@@ -0,0 +1,326 @@
...
+func waitForRollbackWebhook(kubeClient client.Client, logger kitlog.Logger) {
+	Eventually(func() bool {
+		config := &admissionregistrationv1.MutatingWebhookConfiguration{}
+		err := kubeClient.Get(context.TODO(), client.ObjectKey{Name: "theatre-rollback-mutate"}, config)
+		if err != nil {
+			logger.Log("error", err)
+			return false
+		}
+		return true
+	}).Should(Equal(true))
```
0x0013:
```suggestion
	}).Should(BeTrue())
```

Canonical style (already followed elsewhere)

cmd/rollback-manager/acceptance/acceptance.go
```diff
@@ -0,0 +1,326 @@
...
+}
+
+func generateName(prefix string) string {
+	return fmt.Sprintf("%s-%d", prefix, testCounter.Add(1))
+}
+
+func generateTargetName() string {
+	return generateName("target")
+}
```
0x0013: This hard-coded helper doesn't really do anything helpful. I'd suggest just calling `generateName("target")` directly, same as we already do with "previous-target".

cmd/rollback-manager/acceptance/acceptance.go
```diff
@@ -0,0 +1,326 @@
...
+
+func createAnalysisTemplate(kubeClient client.Client, targetName, analysisType, analysisValue string) *analysisv1alpha1.AnalysisTemplate {
+	template := &analysisv1alpha1.AnalysisTemplate{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      generateName(analysisType + "-analysis"),
+			Namespace: namespace,
+			Labels: map[string]string{
+				"acceptance-target": targetName,
+				analysisType:        analysisValue,
```
0x0013: The way Analysis type labelling works is that "true" is the only meaningful value - "false" or any other value would be equivalent to the label being missing.

I would suggest simplifying by only passing `analysisType` in the function, and hardcoding "true" as the value.

Alternatively, if you want the function to be more generic and support Analysis with both health and rollback type, you could accept `[]string` as that argument. But if we don't intend to cover such analysis template/run, it's not necessary.

cmd/rollback-manager/acceptance/acceptance.go
```diff
@@ -0,0 +1,326 @@
...
+			ret = append(ret, rollback)
+		}
+	}
+	return ret
+}
+
+func expectRollbackSucceeded(kubeClient client.Client, name string) {
+	Eventually(func(g Gomega) {
+		rollback := getRollback(kubeClient, name)
+		condition := meta.FindStatusCondition(rollback.Status.Conditions, deployv1alpha1.RollbackConditionSucceded)
```
0x0013: Not directly applicable to the changes in this PR, but there is a pre-existing typo `RollbackConditionSucceded` -> `RollbackConditionSucceeded`

Maybe worth fixing as a drive-by change.

cmd/rollback-manager/acceptance/acceptance.go
0x0013, line 1: There is no cleanup done, such as in an `AfterEach` block, that would remove the resources created for the tests.

This is not a real problem right now, and this comment should not be considered blocking - the atomic `testCounter` ensures there will be no conflicts. However, for correctness, it would still be great to add cleanup, similar to how it's done in `workloads-manager` acceptance tests.

0x0013 (changes requested):
I've added some review comments, but looks good in general.

cmd/rollback-manager/acceptance/acceptance.go
IsaacJames: Implemented now, I opted to add a shared label (`theatre-acceptance-test`) to the all the acceptance test resources so cleaning them up would be a bit easier.

cmd/rollback-manager/acceptance/acceptance.go
IsaacJames: Good point, this was a holdover from the other E2E test suites as you identified, but I agree that it would better live in the `Prepare` function (which already accepts a rest config). I've moved it now.

IsaacJames: On second thoughts, this doesn't actually work since `Prepare` is run in a separate process altogether (`make acceptance-prepare`), so we'll need to go with the `BeforeAll` block.

cmd/rollback-manager/acceptance/acceptance.go
IsaacJames: Agreed

cmd/rollback-manager/acceptance/acceptance.go
IsaacJames: Fair, I've removed the parameter now.

cmd/rollback-manager/acceptance/acceptance.go
IsaacJames: Good catch, I'll include a fix.

cmd/rollback-manager/acceptance/acceptance.go
IsaacJames: Good to know, I've changed this for the Rollback tests now and cut a ticket to update the other theatre tests: https://gocardless.atlassian.net/browse/CI-4056?atlOrigin=eyJpIjoiNzEzYWYxNmFiYTY1NGRjM2FkMzBlOGJjMjU3NjBjMzEiLCJwIjoiaiJ9

cmd/rollback-manager/acceptance/acceptance.go
goelozev: This is marked as resolved, but it's still using the `a` instead of `an`.

IsaacJames: Hmm, I added this suggestion to a batch in the UI and committed it here: https://github.com/gocardless/theatre/pull/506/changes/6a34181f878f21d3ecf3b959545d87324c3e1dea

IsaacJames: Guess I must've overwritten those changes at some point, I've fixed that now.

goelozev (commented):
Looks good! Are you planning to cover also the non-so-happy path of testing whether no rollbacks are created when a previous release was unhealthy?

In regards to your question about under which manager to put these tests under - I think the rollback manager is probably the right place for testing the whole functionality.

goelozev (approved):

PR #519, opened 2026-08-27
Fix: e2e CRD imports for Argo Rollouts CRDs

goelozev:
### Summary

https://github.com/gocardless/theatre/pull/506 added Argo Rollouts CRDs directly in theatre's bundle. If people want to use release analysis, we expect them to install Argo Rollouts themselves, which will include those CRDs.

We should only put those CRDs in the bundle when running our acceptance tests.


0x0013 (approved):
LGTM

PR #522, opened 2026-09-10
Fix DirectoryRoleBinding RoleRef mutability and RoleBinding ownership

0x0013:
## `roleRef` mutability

Previously, `DirectoryRoleBinding` `.spec.roleRef` was mutable, but not properly parsed on change. Thus, a change in roleRef would leave existing and new members of `DirectoryRoleBinding` assigned to the previous `roleRef`.

Changing the RoleRef of an existing `DirectoryRoleBinding` was not intended behavior, thus we resolve this by setting the field as immutable. This mirrors the behavior of native Kubernetes primitives such as `ClusterRoleBinding`.

## `RoleBinding` ownership

DRB reconciliation loop did not check controller ownership of a `RoleBinding`, thus it could mutate a `RoleBinding` resource of the same name, even when the parent `DirectoryRoleBinding` was not the owner of the resource. We now ensure ownership, and fail with "NotOwned" event if an existing resource was not created for this resource.

Jira: FRI-22756

### DIFF
--- a/api/rbac/v1alpha1/directoryrolebinding_types.go
+++ b/api/rbac/v1alpha1/directoryrolebinding_types.go
@@ -8,7 +8,9 @@ import (
 // DirectoryRoleBindingSpec defines the desired state of DirectoryRoleBinding
 type DirectoryRoleBindingSpec struct {
 	Subjects []rbacv1.Subject `json:"subjects"`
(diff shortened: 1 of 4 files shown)


goelozev (approved):

leonard-tesnov (approved):
Seems good to me. It fixes 2 vulns that were identified by the researchers.

PR #523, opened 2026-09-10
Improve acceptance test container framework selection and port conflict

0x0013:
This PR improves the support for different container runtimes.

We previously already provided `CONTAINER_TOOL` variable in Makefile for unverified support. Now we fix several other related issues:

## Image name

Previously we used an override to `theatre:latest` in local acceptance test manifests. This defaults to dockerhub when using Docker, but other runtimes (e.g. Podman) may reconcile it differently.

The result is that the local Kind cluster contains the image of one name (`localhost/theatre:latest`), while it looks for the image of a different name (`docker.io/library/theatre:latest`), and fails to load.

We resolve this by explicitly setting it to a fully qualified name `localhost/theatre:latest`.

Additionally, we define the override as single source of truth where possible, in `config/acceptance/kustomization.yaml` (for overrides that are ingested from it), instead of in each override file individually.

## improve container-tool support

The `CONTAINER_TOOL` selection from Makefile now gets passed to the acceptance test code, to be used directly instead of hardcoding "docker". We also pass it through to Kind for explicit backend support, instead of depending only on auto-detection, which is fragile - if a user was setting `CONTAINER_TOOL` as "podman", but also has Docker installed, auto-detection would choose docker.

## fix orphaned envtest

Some suites using envtest did not properly tear down the test environment on failure.

"deploy" and "runner/integration" were already implemented properly and would clean up, as the cleanup is in `AfterSuite` - we replicate this pattern in other suites depending on envtest.

## metrics port conflict

Not triggered when running `make test`, as it runs `ginkgo -r` which executes sequentially. However, running something like `go test ./rbac/... ./workloads/...` would cause both managers to attempt to launch at the same time, with the default metrics port `:8080`, and thus fail with conflict.

We don't use the metrics server in integration suites, so disable them to prevent this latent bug from appearing later.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,16 +60,16 @@ vet: ## Run go vet against code.
 test: manifests generate fmt vet setup-envtest setup-ginkgo ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" ginkgo -race -randomize-suites -randomize-all -r ./...
 
-acceptance-e2e: install-tools acceptance-prepare acceptance-run acceptance-destroy ## Requires the following binaries: kubectl, kustomize, kind, docker
+acceptance-e2e: install-tools acceptance-prepare acceptance-run acceptance-destroy ## Requires the following binaries: kubectl, kustomize, kind, $(CONTAINER_TOOL)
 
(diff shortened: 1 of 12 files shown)


goelozev (approved):
Nice!

PR #524, opened 2026-09-10
docs(claude): fix

0x0013:
Most of the current `CLAUDE.md` content is a cache of the repo state, or duplication of information found elsewhere:

- repo structure is just cache of `ls`
- API groups is a cache of the CRDs (easily discoverable in the repo)
- commands is basically cache of `make help`
- instructions are mostly duplicated from `README.md`

There are also some misleading version pins which are already out of date (envtest 1.24) - confirmed misleading for agents

The file lacks any real conventions or non-obvious behaviors.

This PR reduces the size by ~80% by removing the likely non-benefitial cached data, and adding non-obvious conventions instead.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,79 +1,18 @@
-# Theatre — CLAUDE.md
+# Theatre
 
-## Project Overview
+Kubernetes operators, admission webhooks, and CLIs. Module path is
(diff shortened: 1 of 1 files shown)


IsaacJames (approved):
Nice!

PR #525, opened 2026-09-23
Add Analysis integration tests

0x0013:
Add integration tests for Analysis functionality

## Preparation

- remove unnecessary uniqueness wrappers for test namespace, leveraging
  `GenerateName` instead
- package-level shared k8sClient, uncached, direct to API server. Reduces
  polling for stale resources
- Gomega Eventually/Consistently timeout and polling duration defaults
- generated targets/commits use randomHex, instead of counter-based uniqueness
- release count is now counted instead of listed by index field. Index
  field lives on the cache, which we no longer use
- refactor existing manager to include optional Analysis support. In current implementation, Analysis as the only mechanism providing automated rollback trigger conditions, thus we do not provision a separate Analysis-disabled manager
- when Analysis is enabled, the Healthy/RollbackRequired fields are owned by the manager, so a manual setting on this field will be reset. Thus, we refactor the previous manual setting, to have the manager set the condition based on (test-provisioned) AnalysisRun resources, emulating production flow.

## Tests covered

Must-have tests not covered by unit tests are implemented.

This covers `AnalysisRun` creation, `ClusterAnalysisTemplate` assignment, application of custom selectors, and related template selection actions.

Jira: CI-3803


internal/controller/deploy/integration/suite_test.go
```diff
@@ -46,11 +48,23 @@ func TestSuite(t *testing.T) {
 var _ = BeforeSuite(func() {
 	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
 
+	// Controllers here react within tens of milliseconds, so poll frequently and
+	// keep the negative-assertion window short.
+	SetDefaultEventuallyTimeout(5 * time.Second)
+	SetDefaultEventuallyPollingInterval(100 * time.Millisecond)
+	SetDefaultConsistentlyDuration(time.Second)
+	SetDefaultConsistentlyPollingInterval(100 * time.Millisecond)
```
goelozev: Nice!

goelozev (approved):

